PR #2, opened 2015-11-18
Fix typo in FSI output window

baronfel:


Krzysztof-Cieslak:
It also has to be changed in package.json in channels definitions

baronfel:
It's fine there

```
"outputChannels": [
	"F# Interactive"
]
```

PR #17, opened 2015-12-01
Mono fix

### LINKED ISSUE #11: Does not build on Linux/Mono
I'm getting a compile error when running build.sh on Linux with Mono 4.2: 

`A namespace and a module named 'Ionide.VSCode' both occur in two parts of this assembly`

If I remove the Mono specific stuff in build.fsx it works fine. Any reason why a separate build process is needed?

azumikav:
this pull request fix #11


Krzysztof-Cieslak:
Nice, thanks :)

Issue #12, opened 2015-11-24
FsAutoComplete.Suave.exe processes left running


hyperlobic:
Under Linux, every time an F# folder is opened an autocomplete process is started up. These processes are never killed, even if the application is closed.

azumikav:
The go extension has this problem too https://github.com/Microsoft/vscode-go/issues/26

Krzysztof-Cieslak:
Should be fixed in `1.2.2`

PR #18, opened 2015-12-02
Kill FsAutoComplete.Suave.exe when vscode closes

azumikav:
Right now I dont have windows available to test if it also works on windows.
#12


azumikav:
nevermind, this is still bad.

PR #41, opened 2016-03-04
Add syntax for `inline` so that inlined methods and non-inlined

rojepp:
methods highlight in the same way


Krzysztof-Cieslak:
Probably same should be PRed to `ionide-fsgrammar` repo for Atom highlighting (and hopefully VSCode in future)

PR #42, opened 2016-03-05
handle filePath with spaces in url

falfaddaghi:
I was not able to build/run the fake file due to a git issue on my pc, but I am fairly sure this will fix the issue. if it does not i will keep trying to fix the git issue.

### DIFF
--- a/src/Components/Linter.fs
+++ b/src/Components/Linter.fs
@@ -7,6 +7,7 @@ open FunScript.TypeScript.vscode
 open FunScript.TypeScript.vscode.languages
 open FunScript.TypeScript.path
 open FunScript.TypeScript.fs
+open System.Net
 
 open DTO 
 open Ionide.VSCode.Helpers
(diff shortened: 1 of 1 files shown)


Krzysztof-Cieslak:
Git issue is probably SSH key issue, following https://help.github.com/articles/generating-an-ssh-key/ should be enough to get it working

rojepp:
I was going to ask about that. Why not use https for the git deps? It's what I did locally to make it build. Would probably make it easier for would-be contributors.

Krzysztof-Cieslak:
That's reasonable suggestion. I'll change that in some free time... or we accept PRs ;)

falfaddaghi:
I changed paket to use https and how I get this error. Pretty Ironic.
I will move the code to a different workspace and push the pr if it works.

> Message: Could not run "git clone file:///C:\Users\FAISAL ALFADDAGHI\.paket\git\db\FSharpFormatting.CLI".
Error: 
Cloning into 'ALFADDAGHI\.paket\git\db\FSharpFormatting.CLI'...
fatal: 'C:/Users/FAISAL' does not appear to be a git repository

falfaddaghi:
got the project building and my fix is away off i will keep trying .
I am going to close this pr

PR #44, opened 2016-03-05
changed paket to use https

falfaddaghi:


falfaddaghi:
seems like you already merged a pr for this.

Issue #49, opened 2016-03-29
Ionide fails to register if no project open


CrshOverride:
Currently, `ionide` fails to register with VSCode on OSX if there is no current project open.

To Reproduce:

1. Install a fresh copy of VSCode.
2. Immediately installe `ionide` via `ext install ionide-fsharp`.
3. Restart
4. Attempt `fsi start`. You should get a message that nothing is registered to handle `fsi.start`. If you open the developer tools, you'll see console errors about `path expects a string`.

PR #50, opened 2016-03-29
Allow Ionide in VSCode w/ No Project Open

CrshOverride:
This PR addresses #49. When `workspace.Globals.rootPath` is `null`, we simply return an empty list so that nothing is processed. This allows registration to continue and the plugin to function as expected.


Krzysztof-Cieslak:
Thanks! Will be released in next version. :)

PR #62, opened 2016-05-16
Export the deactivation function

hyperlobic:
The deactivation function was not being called, resulting in the language service not getting cleaned up.


Krzysztof-Cieslak:
Thanks! :)

Issue #4, opened 2015-11-18
Include FSI in package


xavierzwirtz:
Getting an exception when I try to eval `let x = ""`

```
[Plugin Host] This socket is closed.
workbench.main.js:1602 [Plugin Host] Object {}
workbench.main.js:1602 [Plugin Host] {"Kind":"error","Data":"File 'C:\\3' not parsed"}
workbench.main.js:1602 [Plugin Host] unhandled rejected promise Object {}(anonymous function) @ workbench.main.js:1602emitTwo @ events.js:87emit @ events.js:172handleMessage @ internal/child_process.js:685channel.onread @ internal/child_process.js:440
workbench.main.js:1626 errore.onUnexpectedError @ workbench.main.js:1626(anonymous function) @ workbench.main.js:1626e.onUnexpectedError @ workbench.main.js:1548u @ workbench.main.js:1547e.onUnexpectedPluginHostError @ workbench.main.js:1604e.handle @ workbench.main.js:1588a @ workbench.main.js:1576h.handle @ workbench.main.js:1576(anonymous function) @ workbench.main.js:1602(anonymous function) @ workbench.main.js:1602emitTwo @ events.js:87emit @ events.js:172handleMessage @ internal/child_process.js:685channel.onread @ internal/child_process.js:440
workbench.main.js:1602 [Plugin Host] [E] 2015-11-18T21:26:55.9116471Z: request failed [Suave.Web.defaultErrorHandler] exn:
Microsoft.FSharp.Compiler.ErrorLogger+WrappedError: Exception of type 'Microsoft.FSharp.Compiler.ErrorLogger+WrappedError' was thrown.
   at Microsoft.FSharp.Compiler.Lexhelp.reusingLexbufForParsing[a](LexBuffer`1 lexbuf, FSharpFunc`2 f)
   at Microsoft.FSharp.Compiler.Lexhelp.usingLexbufForParsing[a](LexBuffer`1 lexbuf, String filename, FSharpFunc`2 f)
   at Microsoft.FSharp.Compiler.SourceCodeServices.Parser.ParseOneFile(String source, Boolean matchBracesOnly, Boolean reportErrors, String mainInputFileName, FSharpList`1 projectSourceFiles, TcConfig tcConfig)
   at <StartupCode$FSharp-Compiler-Service>.$Service.ParseAndCheckFileInProject@2470.Invoke(CancellationToken _ct)
[...]

PR #94, opened 2016-07-13
Try to find Fsi.exe in typical install dir if not on path

oxxa:
Use VSCode.getFsiFullPathWin (submitted in another PR to ionide-vscode-helpers) so that users who do not have Fsi.exe on their path should still get a working interpreter out of the box.

(Also some changes I need to make to ge the build working on Windows)


oxxa:
This could also help to close #4

oxxa:
Closed while the matching PR in Helpers can be reworked

PR #98, opened 2016-07-20
Look for fsi in common install paths on Windows

oxxa:
Attempt to get interactive working out of the box even if the Fsharp install dir is not on path on windows.

Added Environment.fs which contains code to look for the fsi.exe in the standard install paths on Windows (note for Linux/Mac I just assume it is on path).  This path is then used on spawning fsi in Fsi.fs.  I have not been able to test this on non-windows but as lot of the code is adapted fromthe FSAutocomplete project it should be robust.
(Also this commit adds a change to paket.dependencies to use standard https for fsgrammar.)

### DIFF
--- a/src/Components/Environment.fs
+++ b/src/Components/Environment.fs
@@ -1,0 +1,42 @@
+namespace Ionide.VSCode.FSharp
+
+//---------------------------------------------------
+//Find path of F# install and FSI path
+//---------------------------------------------------
+// Below code adapted from similar in FSSutoComplete project
+module Environment = 
+    open Fable.Core
+    open Fable.Import.Node
+    open Fable.Import.Node.fs_types
+    
+    let isWin = ``process``.platform = "win32"
+    
+    let (</>) a b = 
+        if isWin then a + @"\" + b
+        else a + "/" + b
+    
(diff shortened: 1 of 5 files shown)


src/Components/Environment.fs
```diff
@@ -0,0 +1,42 @@
...
+            if detected = null then @"C:\Program Files (x86)\"
+            else detected
+    
+    let private fsharpInstallationPath = 
+        [ "4.0"; "3.1"; "3.0" ]
+        |> List.map (fun v -> programFilesX86 </> @"\Microsoft SDKs\F#\" </> v </> @"\Framework\v4.0")
+        |> List.tryFind dirExists
+    
+    let fsi = 
+        if true then "fsharpi"
```
Krzysztof-Cieslak: I guess it should be `if not isWin` ;)

oxxa:
Argh!  I put that in to test how it would behave on linux, apologies.  Will fix asap

Krzysztof-Cieslak:
Also, please rebase, if possible. I've just pushed some changes to master

Krzysztof-Cieslak:
Thanks a lot! :)

Krzysztof-Cieslak:
Released in `2.2.0`

Issue #99, opened 2016-07-21
Issues launching homebrew Mono on Mac OS X El Capitan


mariosangiorgio:
It looks to be the very same issue experienced by [OmniSharp](https://github.com/OmniSharp/omnisharp-atom/issues/42).
The only way I can get Ionide to work properly is to launch Visual Studio code from the terminal (which has `mono` in the path).

It's an issue related to Mono installed via Homebrew. The OmniSharp team solved it [in this way](https://github.com/OmniSharp/omnisharp-sublime/commit/9b9b85cff954550c85109b65733e067762bb0366).

To be honest it doesn't look too clean so I was wondering if there is another way to get everything to work as expected.

Krzysztof-Cieslak:
Mono path can now be provide in settings using `FSharp.monoPath` option

PR #103, opened 2016-07-21
Add configuration option to specify tools dir

oxxa:
**Search for `fsi ` first in the path specified in the user settings file, then if not found (and on windows) try the standard install dirs, then if not found just use the exe name and presume it will be found on path.**

Just revisting my previous commit in light of #99 I thought it would be good to have a configuration option for the tools path.  This should help those people who can't/don't want to add the tools dir to their path.  I saw it was also suggested to include fsi.exe with the extension ( #4 ), or offer to download it for the user, but until then this should work ok.

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -155,15 +155,20 @@
       }
     ],
     "configuration": {
-        "type": "object",
-        "title": "FSharp configuration",
-        "properties": {
-            "FSharp.automaticProjectModification": {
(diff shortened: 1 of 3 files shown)


Krzysztof-Cieslak:
Great, thanks a lot again! :)

Krzysztof-Cieslak:
Released in `2.2.6`

PR #129, opened 2016-08-10
Log FSAC http request failures to console,

adamchester:
regardless of the value of `FSharp.logLanguageServiceRequestsToConsole`, because errors should be seen, and shouldn't happen often (if all goes well).

screen shot 2016-08-10 at 12 44 14


cloudRoutine:
good improvement, I'm always for more accurate error info :+1:

PR #130, opened 2016-08-10
Don't init services until FSAC is listening

### LINKED ISSUE #128: Language service sometimes not ready for first project load request
Turns out we still have the issue after this change: #110 

Even though FSAC.suave has been launched, it hasn't necessarily started accepting/listening yet.

I'll look into this further.

adamchester:
This fixes #128.


cloudRoutine:
LGTM

PR #154, opened 2016-09-18
Added Nancy Selfhosted template support

2moveit:
Is there a reason not to use the template.json from Forge or just parse the folders?

---
Created at #remmidemmiconf


Krzysztof-Cieslak:
@2moveit, Thanks!

No, no particular reason.

I'd welcome PR adding parsing template.json and using it instead ;)

PR #163, opened 2016-09-26
Add CopyFSAC for Default target

kirill-gerasimenko:
Hi,

On the clean ionide clone if run F5 in vscode - it fails with the "Failed to start language services. Please check if Microsoft Build Tools 2013 are installed". This is because FSAC is not copied to the release folder on default target in build.fsx. 
This PR fixes this.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -229,6 +229,7 @@ Target "Test" DoNothing
 
 "Clean"
 ==> "RunScript"
+==> "CopyFSAC"
 ==> "Default"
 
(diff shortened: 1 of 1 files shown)


Krzysztof-Cieslak:
I think I'll prefer it as is right now - I'm using symlinked version of plugin so always latest version is load, which locks FSAC files. So trying to run `CopyFSAC` from VSCode results in failed build (as used file cannot be overwritten).

kirill-gerasimenko:
No problem

PR #177, opened 2016-09-30
Fsi extra params

kirill-gerasimenko:
Hi,

While I was working on some scripts I referenced FSharp.Core directly, but FSI doesn't start if you do so, it wants you to have "--noframework" parameter passed during the start in this situation. 
So this PR adds ability to provide extra parameters to FSI startup from separate setting. 

I hope it will be useful not only to me.

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -215,6 +215,11 @@
           "type": "boolean",
           "default": true,
           "description": "Enables integration with FSharpLinter (additional warnigns)"
+        },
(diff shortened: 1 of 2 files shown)


Krzysztof-Cieslak:
Thanks! :)

kirill-gerasimenko:
You're welcome :) Now I can use my scripts, hooray :)

PR #182, opened 2016-10-05
Correct spelling of warnings

chamook:
Figured this would be irritating to get an issue for when it was so easy to fix. This just corrects the spelling of warnings in the config description for FSharp.Linter.


Krzysztof-Cieslak:
Thanks!

PR #203, opened 2016-10-13
Do not show code lens for abstract members

vasily-kirichenko:
+ show code lens for val bindings. 

https://github.com/fsharp/FsAutoComplete/pull/129 must be merged and referenced by this PR before it's merged.

### DIFF
--- a/paket.lock
+++ b/paket.lock
@@ -1,7 +1,7 @@
 FRAMEWORK: NET45
 NUGET
   remote: https://www.nuget.org/api/v2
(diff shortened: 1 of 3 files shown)


vasily-kirichenko:
@Krzysztof-Cieslak why do you handle "Extension method" in a special way here https://github.com/ionide/ionide-vscode-fsharp/blob/master/src/Components/CodeLens.fs#L26-L32? I've removed it and now code lens shows OK for `default` and `override` members:

image

PR #216, opened 2016-10-19
use plane text for comment section in tooltips

vasily-kirichenko:
image


Krzysztof-Cieslak:
Released in `2.7.1`.

Thanks!

PR #218, opened 2016-10-20
Better tooltips

vasily-kirichenko:
image

### DIFF
--- a/src/Components/Tooltip.fs
+++ b/src/Components/Tooltip.fs
@@ -25,18 +25,27 @@ module Tooltip =
                             "language" ==> lang
                             "value" ==> value.Trim()
                         ] |> Case2
+                    
+                    let fsharpBlock (lines: string[]) : MarkedString =
+                        lines |> String.concat "\n" |> markStr "fsharp"
+
                     let sigContent =
-                        res.Signature.Split('\n')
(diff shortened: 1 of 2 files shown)


vasily-kirichenko:
I also added replacement `<c>xxx</c>` in autocomplete. It does not support Markdown, but backticking is better then xml tags I think.

image

cloudRoutine:
iirc the text content of the autocomplete panes can also be styled via markdown

cloudRoutine:
nope i was wrong markedstrings are only for hovers and decorations

Krzysztof-Cieslak:
That sounds like something what should be fixed in Code.

Krzysztof-Cieslak:
@cloudRoutine will you create issue on their repo?

cloudRoutine:
someone beat us to it https://github.com/Microsoft/vscode/issues/11877

EmmanuelOga:
@vasily-kirichenko: Tangential question, any idea what it would take to provide better tooltips like this on Visual Studio?

Txs!

vasily-kirichenko:
@EmmanuelOga it's possible. I think MS team is working on it right now (Roslyn-style/based tooltips in VS 2017).

PR #222, opened 2016-10-20
pass file version to FSAC

vasily-kirichenko:


vasily-kirichenko:
https://github.com/ionide/FsAutoComplete/pull/16

PR #225, opened 2016-10-20
Remove <para></para> in Markdown.replaceXml

kirill-gerasimenko:
Hi,
This is tiny PR to remove `<para>` from comments, so this

image

looks like this

image


Krzysztof-Cieslak:
Thanks!

PR #233, opened 2016-10-22
new tooltips

vasily-kirichenko:
WIP

Requires https://github.com/ionide/FsAutoComplete/pull/19


vasily-kirichenko:
It does not show interface members:

image

old:

image

vasily-kirichenko:
OK, I'm not sure we should merge this since it's not 100% better than the current tooltips. 

I think we should try to fix all the issues, than consider if it's worth to use it.

Issue #220, opened 2016-10-20
Lint not working


csmager:
This is initially more a question than a bug, as I could well be doing something wrong. I've searched, and I can't find any documentation or explanation as to how this is supposed to work.

The default setting for `FSharp.linter` is `true`, so I'd assume it is enabled.  I'm not seeing any UI that gives me the feedback I see in F# Power Tools in VS 2015. I've added an `Settings.FSharpLint` file in case this was required - again, F# Power Tools in VS 2015 picks this up.

I turned on the debug logging and see entries like this:

>[10:45:46 INFO ] REQ (022) -> {lint}, File = "~\src\abc\def.fs"
[10:45:46 DEBUG] RES (022) <- {lint} in 4 ms: Kind={"error"}
Data="File 'c:\\path\\abc\def.fs' not parsed"

Krzysztof-Cieslak:
Are any other features (autocomplete, error highlighting, tooltips etc) working? 
Can you show full (from start of plugin) log?

csmager:
No issues with anything else - I get auto complete, see CodeLens, the type & docs tool tips, error highlights / tool tips.  I'm using v2.7.1, but this has never worked for me (I've only recently realised it's supposed to be enabled by default).

I created a simple console app, here's the full log from opening folder and clicking `Program.cs`:

>[11:25:24 DEBUG] got FSAC line, is it the started message? true
[11:25:24 DEBUG] FSAC stdout: [I] 2016-10-20T10:25:24.2054001Z: listener started in 17.533 ms with binding 127.0.0.1:8815 [Suave.Tcp.tcpIpServer]
[11:25:24 INFO ] REQ (001) -> {project}, File = "~\IonideTest\IonideTest.fsproj"
[11:25:24 INFO ] REQ (002) -> {lint}, File = "~\IonideTest\Program.fs"
[11:25:24 INFO ] REQ (003) -> {parse}, File = "~\IonideTest\Program.fs"
[11:25:24 DEBUG] RES (002) <- {lint} in 311 ms: Kind={"error"}
Data="File 'c:\\dev\\IonideTest\\IonideTest\\Program.fs' not parsed"
[11:25:24 DEBUG] RES (001) <- {project} in 575 ms: Kind={"project"}
[...]

PR #239, opened 2016-10-26
Linter: Check event is defined before accessing property

duckmatt:
Small change to fix error in https://github.com/ionide/ionide-vscode-fsharp/issues/220


duckmatt:
Closed - going to resubmit with a commit under my account

PR #240, opened 2016-10-26
Linter: Check event defined before accessing property

duckmatt:
Small change to fix error in #220

Sorry for previous PR, accidentally committed on wrong account


Krzysztof-Cieslak:
Thanks!

Issue #238, opened 2016-10-26
Send to FSI doesn't make terminal session visible (if hidden)


piers7:
If terminal window is not running at all, 'send to FSI' brings up terminal window focused on FSI. However:

- if FSI previously launched and terminal pane closed (but terminal session not killed - ie fsi still running), send to FSI doesn't reopen terminal.
- if terminal open, but on a different session (say I'm running a console session and an FSI session), send to FSI won't bring the terminal back to the FSI session.

Both can be very confusing - I though that FSI integration in VS Code had just gone back to not working for me, whereas in fact I'd just executed the thing I needed done several time in frustration...

Krzysztof-Cieslak:
Hmm...

I wonder if @kirill-gerasimenko  can take a look at it.

kirill-gerasimenko:
Hi,

Yep, I can look at it :)

kirill-gerasimenko:
That one was easy :)

Krzysztof-Cieslak:
Fix released in `2.8.3`. Thanks @kirill-gerasimenko !

kirill-gerasimenko:
Np.

It looks like the terminal is focused after you send text there. I haven't noticed that during testing. Let me check if parameters to terminal.show fixes it.

Krzysztof-Cieslak:
I think there is parameter for that in `terminal.show`

kirill-gerasimenko:
Yes, will PR it now.

PR #245, opened 2016-10-27
Preserve focus in current window, so it doesn't jump to terminal

kirill-gerasimenko:
Amendment to #238


Krzysztof-Cieslak:
Thanks!

PR #259, opened 2016-10-30
make CodeLens cache per-file

vasily-kirichenko:


Krzysztof-Cieslak:
Already pushed fix. But thanks anyway!

PR #260, opened 2016-10-30
store CodeLens cache in a map

vasily-kirichenko:


cloudRoutine:
Is the resultant JS via Fable different for a Map than for a Dictionary or do they end up the same?

PR #271, opened 2016-11-06
perform an npm update to make sure we've got the latest fable 

baronfel:
This PR makes use of npm's script conventions to make sure that we've got the right dependencies in place before build. This should make newcomers' experience more smooth, as well as help to get people who haven't updated in a while running!

To make this work I locked down Fable's dependencies to the 0.6 version line, which we will now need to explicitly update from now on.

In addition, later we may want to move to yarn to get the benefits of a lockfile and a reproducible build.

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -272,17 +272,18 @@
     "vscode.fsharp"
   ],
   "scripts": {
+    "prebuild":"npm update",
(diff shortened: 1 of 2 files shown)


Krzysztof-Cieslak:
Thanks

Issue #276, opened 2016-11-13
Feature Request: New FSI Command: `Send Last Selection`


fbehrens:
Hi, 
thanks for this great plugin, to begin with.

This feature would be very useful to me in `highly interactive` (tm) f# development, because:

When I'm working on some function, either I like to have some code in another window which reloads the file, opens and calls the function with some value. (Or calls a specific fuchu testcase).

With such functionality I can stay in the implementation file, and call my function with one keystroke to `FSI: Send Last Selection`. 

Without the feature I have to 
* change editor window
* select code
* `FSI: Send selection`
* switch Editor window back back.

Just want to ask if you would accecpt a merge request on that feature ?

Krzysztof-Cieslak:
Sounds good for me.

PR #284, opened 2016-11-18
save current file on SendLastSelection

fbehrens:


rojepp:
Why would you want to do this? 
I have my vscode set to save all the time, but if a user doesn't have it like that, this change seems kind of inappropriate?

fbehrens:
Please see #276. When I am implementing a function, i am calling the function on after ever incremental small change from `fsi`.  This is usually something like 

* loading the file
* opening the namespace
* call the function with desired parameters

With this send last selection it is convenient to have that code in another editor window,
select it once, and than can call it without leaving the implementation file. 
For that it is also required to save the before loading it. 
For that I like to have this saving the file, so I can edit my implementation file, and on one keystroke I can run it.

Yes, I see that it might be a little unexpected, and I see I can use `autoSave` 

For me this is a very productive feature.
I have mapped it to `shift+cmd+s` and I use it in the sense of `save and run my code`

Krzysztof-Cieslak:
I think it should be opt-out, we can use config for that.

fbehrens:
Thanks, i like this.

I have added the configuration option.

There is one problem with this implementation. As the command `workbench.action.files.save` is asynchronous, the `send` happens before the `save` . I tried but I haven't found out how to wait for the `save` to have happened. Maybe s.o. has a hint :)

src/Components/Fsi.fs
```diff
@@ -105,8 +105,10 @@ module Fsi =
     let private sendLastSelection () =
         match lastSelectionSent with
         | Some x ->
-            send x
-            |> Promise.suppress // prevent unhandled promise exception
+            if "FSharp.saveOnSendLastSelection" |> Configuration.get false then
+                commands.executeCommand "workbench.action.files.save" |> ignore
```
Krzysztof-Cieslak: `commands.executeCommand` (should) returns Promise. So you can do `|> Promise.bind(fun _ -> send x)`

fbehrens:
Thanks for the hint. I had to replace `commands.executeCommand "workbench.action.files.save"` with `window.activeTextEditor.document.save ()` to have it now working as expected.

 I squached the commits and would appreciate a merge.

Krzysztof-Cieslak:
Released in `2.10.1`

PR #293, opened 2016-11-28
alt+Enter keybinding only in fsharp editor

### LINKED ISSUE #292: The "send selection" shortcut should not be global
Hello,

Currently the "send selection" (to fsi) shortcut is available in contexts other than working with F# and it's annoying, because Alt+Enter could very well be used for other programming languages and their REPLs. For example I'm also playing with OCaml and I need `ocaml.utop_send`.

The right way to define this shortcut is to add an `editorLangId == 'fsharp'`, like so:

```javascript
{ 
  "key": "alt+enter",             
  "command": "fsi.SendSelection",
  "when": "editorFocus && !findWidgetVisible && editorLangId == 'fsharp'" 
}
```

Thanks,

slang25:
Fixes #292

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -163,7 +163,8 @@
       {
         "command": "fsi.SendLine",
         "key": "alt+/",
-        "mac": "alt+/"
+        "mac": "alt+/",
+        "when": "editorFocus && !findWidgetVisible && editorLangId == 'fsharp'"
       },
       {
         "command": "fsi.SendSelection",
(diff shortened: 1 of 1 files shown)


Krzysztof-Cieslak:
This should probably be also added for `sendLine`

slang25:
Good suggestion. I was tempted for `SendLine` to add `&& !editorHasSelection`, but wasn't sure so I've left it off.

Krzysztof-Cieslak:
Thanks!

PR #335, opened 2017-01-15
A separate OutputChannel for FSAC Stdout lines in DEBUG

adamchester:
Recently, I had a number of crashes inside `FsAutoComplete` and it was too difficult to setup the appropriate logging to dive in and get a clue about why. 

If this PR is merged, and when logging is configured with `DEBUG` level and `Console` (`OutputChannel`) enabled, all the output from `FsAutoComplete` is captured into a separate `OutputChannel`.

Additionally, there are potentially many concurrent requests to the language service, and it was difficult to correlate the logs between client (vscode) and server (FsAutoComplete). To address that, I have added the `?requestId=` query string to every request sent to the language service.

Separately from this PR, we can easily add additional logging inside the `FsAutoComplete.Suave` project or `FsAutoComplete` itself. However, for now, the only thing visible in this new output channel is as follows:
screen shot 2017-01-15 at 15 28 26

### DIFF
--- a/src/Core/LanguageService.fs
+++ b/src/Core/LanguageService.fs
@@ -38,21 +38,37 @@ module LanguageService =
         with
         | _ -> Level.INFO
 
-    // Always log to the logger, and let it decide where/if to write the message
-    let log =
+    // note: always log to the loggers, and let it decide where/if to write the message
+    let createConfiguredLoggers source channelName =
+
         let channel, logRequestsToConsole =
             match logLanguageServiceRequestsConfigSetting with
             | LogConfigSetting.None -> None, false
-            | LogConfigSetting.Both -> Some (window.createOutputChannel "F# Language Service"), true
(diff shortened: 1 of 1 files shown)


src/Core/LanguageService.fs
```diff
@@ -250,23 +266,23 @@ module LanguageService =
                 let outputString = n.ToString()
                 // Wait until FsAC sends the 'listener started' magic string until
                 // we inform the caller that it's ready to accept requests.
-                let isStartedMessage = outputString.Contains ": listener started in"
+                let isStartedMessage = outputString.Contains "listener started in"
```
adamchester: This makes the startup detection work for Suave 2.0+ (and still works now too)

Krzysztof-Cieslak:
Amazing, thanks.

Will you also PR logger to FSAC?

adamchester:
Yep, will do 👍

Krzysztof-Cieslak:
Released in `2.19.3`

Issue #142, opened 2016-08-21
Build/ Rebuild project


isaacabraham:
I'd like to add support for build / rebuild project to the .proj file in the current folder, I guess using MSBuild. What's the best way you'd recommend to add this?

Krzysztof-Cieslak:
1. You need to find msbuild ( I guess something like checking usual installation paths [we do it in case of FSI], and maybe providing setting for specifying path to msbuild), it's much easier in case of mono as xbuild is usually in PATH I think. 

2. Invoke msbuild / xbuild with right parameters (Debug vs Release; Build vs Rebuild vs Clean [and maybe vs CustomTarget])

3. Somehow pick good fsproj file (currently open, give user list of all projects in repo and pick one, or something like that). I guess I would go with 2 options - picking from list, and building "default" (relative path to "default" specified in settings so people can override it on workspace base).

4. What about output path? (We may want to add option to specify it [through settings file]... or do nothing, so path specified in fsproj is used [that's how VS does it])

isaacabraham:
I would just want to start with the simplest thing - look at the current file that's open, find the project that's either in the same folder or upwards of it, and do a debug build. In the future the configuration could be parameterised - you can have multiple configurations (not just debug and release), and I don't want to go through that whole hassle for v1 of this. Same for output path - just do the standard build of the project, the same as doing a project build in Visual Studio.

So, questions for moving forward on this: -

1. Where should settings be stored (for e.g. specifying MSBuild path)?
2. Where should this feature live in terms of this repo (more or less)?
3. What's the easiest way to test this out locally? I don't want a repeat of my PR to the Ionide Paket repo where I needed 4 releases to get it right!

cloudRoutine:
it'd be in this one, the msbuild property info you need is extracted via forge. MsBuild has to be in a standard location otherwise everything else will get broken across ms programming tools. You also need to check the tools version listed in the project files, there are going to be multiple versions of msbuild installed on any system so you need to make sure you use a compatible one. Configurations live both in the solution and the project files, so you need to check the solution file to make sure that if you're trying to build multiple projects under the same configuration settings that it's actually viable. 

I think looking at the file that's currently open and trying to infer the project it's a part of is a bad approach. If they're added contextual menu items (it's coming soon if it's not in already) to the api right clicking on the project file in the file tree and adding a "Build Project" action to the context menu is more reasonable. 

[...]

PR #338, opened 2017-01-18
add initial msbuild tasks and configuration

baronfel:
Very basic initial implementation of #142. This doesn't do anything complex like specifying targets or configurations, but it does work on my local projects just fine.

### DIFF
--- a/src/Core/Utils.fs
+++ b/src/Core/Utils.fs
@@ -56,8 +56,11 @@ module Option =
 
 [<RequireQualifiedAccess>]
 module Document =
-    let (|FSharp|Other|) (document : TextDocument) =
-        if document.languageId = "fsharp" then FSharp else Other
+    let (|FSharp|CSharp|VB|Other|) (document : TextDocument) =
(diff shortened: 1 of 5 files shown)


src/Core/Utils.fs
```diff
@@ -56,8 +56,11 @@ module Option =
 
 [<RequireQualifiedAccess>]
 module Document =
-    let (|FSharp|Other|) (document : TextDocument) =
-        if document.languageId = "fsharp" then FSharp else Other
+    let (|FSharp|CSharp|VB|Other|) (document : TextDocument) =
```
baronfel: Expanded this to include other msbuild-based languages, but it may be worth splitting msbuild out into a language-agnostic component?

Krzysztof-Cieslak: I don't think I care enough to create separate plugin for it. ;)

(It may change with introducing windows debugger since it also works for C# and VB)

Krzysztof-Cieslak:
Thanks a lot. :)

PR #390, opened 2017-03-11
on non-windows systems, probe for msbuild/xbuild wrappers

baronfel:
The msbuild functionality wasn't working at all on non-windows systems for two reasons:

1. in `Environment.fs` we were defaulting to `xbuild`, which is fine except we have to use the shell to resolve that binary out to a full path before we can use it, and
1. the mechanism that we would normally use to get around this, the `FSharp.msbuildLocation` config setting, was only being consulted if the `Environment.msbuild` value was `None`, which was never the case on non-windows systems.

This PR does two things:

1. Addresses point 1 by probing the for msbuild/xbuild so that we have sane defaults, and 
1. Addresses point 2 by changing the order of msbuild resolution so that we always consult the user configuration before we do any probing at all.  This should help head off any potential errors and give users a way out in case of the world exploding.

This also helps ensure that various tools that want to use msbuild (eg the Expecto integration) are resolving the same way all the time, instead of what they were doing before setting disparate default values.

### DIFF
--- a/src/Components/Environment.fs
+++ b/src/Components/Environment.fs
@@ -10,6 +10,7 @@ module Environment =
     open Fable.Import.vscode
     open Fable.Import.Node
     open Fable.Import.Node.fs_types
+    open Ionide.VSCode.Helpers
 
     let isWin = ``process``.platform = "win32"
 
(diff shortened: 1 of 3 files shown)


Krzysztof-Cieslak:
Thanks a lot!

baronfel:
NP, I was working on some stuff for work and realized that nothing ever was working :(

Issue #423, opened 2017-05-01
MsBuild commands should run dotnet CLI for new fsproj


Krzysztof-Cieslak:
Same behavior was implemented for Expecto commands in https://github.com/ionide/ionide-vscode-fsharp/pull/413

@baronfel, would you be interested adding this or should we leave it for some other volunteers? :-)

baronfel:
I'll take a stab, but I won't be able to look until tomorrow or Wednesday :)

enricosada:
i implemented that as part of #455 will send a separate PR.
added also a command `MSBuild: switch msbuild host` to change the current msbuild host (.net `msbuild.exe` or .net core `dotnet msbuild`)

PR #424, opened 2017-05-01
WIP: netcore MSBuild commands

baronfel:
got bored, this is most of the core logic for #423.  Can't build/test from my laptop due to mono issue that I'm investigating.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -3,7 +3,7 @@
 	"version": "0.2.0",
 	"configurations": [
 		{
-            //"preLaunchTask": "Default",
+      "preLaunchTask": "Default",
 			"name": "Launch Extension",
 			"type": "extensionHost",
 			"runtimeExecutable": "${execPath}",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -3,7 +3,7 @@
 	"version": "0.2.0",
 	"configurations": [
 		{
-            "preLaunchTask": "Default",
+            //"preLaunchTask": "Default",
 			"name": "Launch Extension",
 			"type": "extensionHost",
 			"runtimeExecutable": "${execPath}",
--- a/paket.lock
+++ b/paket.lock
@@ -1,7 +1,7 @@
 FRAMEWORK: NET45
 NUGET
   remote: https://www.nuget.org/api/v2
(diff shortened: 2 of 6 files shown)


baronfel:
FAKE update fixed it. Having problems with FsAutoComplete 'project' command working when I run the Suave endpoint standalone and send files to it, but not when run from the instance of FsAutoComplete that's started by ionide.

In addition, this PR doesn't successfully build projects that use the new fsproj but target the full-framework, because of the need to specify the FrameworkIdentifier that @enricosada noted in the wiki: https://github.com/dotnet/netcorecli-fsc/wiki/.NET-Core-SDK-rc4#using-net-framework-as-targets-framework-the-osxunix-build-fails

We should possibly extend FSAC to specify the runtime/TFS to build a project for as part of the project request?

enricosada:
@baronfel yes, that's the only way.

I'll workaround in FSAC with https://github.com/fsharp/FsAutoComplete/issues/162 to at least build the **first** target framework specific in `TargetFrameworks`.

but is just a workaround. ionide need to pass a choose a target frameowkr., It's also really really a good idea to add a way to choose the first project or the sln, because otherwise is random (based on first file open, but with old and new in parallel doesnt work)

enricosada:
@baronfel but why add msbuild commands? 

vscode build (CTRL-SHIFT-B) works afaik, just choose .NET Core or msbuild. you can add more targets there 
 ( @Krzysztof-Cieslak that's how vscode works right? launch and build.json)
Shouldnt we add more build target there?

Also msbuild command can be run from terminal afaik.

Also a note. new fsproj DOESNT require .net code sdk (the `dotnet` command). doesnt need to run `dotnet build` or `dotnet msbuild`.
will work normally from msbuild.exe or mono msbuild (mono 5.0).

baronfel:
The commands here are meant to be super simple happy path helpers to quickly compile a single project, not necessarily run your whole solution build. We can definitely use the normal task infrastructure bit that would require more fiddling to detect and build a particular project I think.

enricosada:
Just set the directory or fsproj in the "args" of the "build" command. 
Afaik helper should set that

Krzysztof-Cieslak:
This will be handled bit differently with #460

PR #448, opened 2017-06-09
Don't require space after "(" or "[" to close pair

inosik:
This makes the brace completion behave more like in Visual Studio (can only speak of VS 2013). It won't auto-close to `[||]` after <kbd>[</kbd> <kbd>|</kbd>, for example.


Krzysztof-Cieslak:
Thanks a lot!

PR #460, opened 2017-06-30
Support .net core msbuild

### LINKED ISSUE #423: MsBuild commands should run dotnet CLI for new fsproj

enricosada:
Currently just the .net msbuild (`msbuild.exe` on win, and `msbuild/xbuild` on
non win) is supported

This add .NET Core msbuild (dotnet msbuild), used by .net core cli (the
`dotnet build` = `dotnet msbuild /t:Build`)

The `MSBuild: ...` commands remains the same, but is possibile to switch between
`.net <-> .net core` msbuild to use with the new `MSBuild: switch msbuild host`
command

Others:

- add `Restore` target
- msbuild commands now return a `Promise` instead of `unit` so is possibile to
  execute and continue, like:

  ```fsharp
  vscode.commands.executeCommand("MSBuild.restore", projectFullPath, 2)
  |> Promise.bind (fun exitCode -> ... )
  ```

fix #423 

TODO:

- [x] default host: atm default host is `None`
     - [x] add configuration `[ ".net"; ".net core"; "ask at first usage" ]`
     - [x] use a sane default (probably `.net msbuild` now for backward compatibility)
- [x] check all msbuild commands return a promise
- [x] create a type of argument to pass optionally to command to specify host/target, so can be run programmatically (not by user choice)

Enable additional work:

- switch default on `autodetect`, and use host based on the fsproj sdk. This require to know what type is the fsproj, so need new info returned from FSAC

### DIFF
--- a/src/Components/MSBuild.fs
+++ b/src/Components/MSBuild.fs
@@ -74,7 +74,7 @@ module MSBuild =
             logger.Debug "No MSBuild host selected yet"
             let cfg = vscode.workspace.getConfiguration()
             let p =
-                match cfg.get ("FSharp.msbuildHost", "auto") with
+                match cfg.get ("FSharp.msbuildHost", ".net") with
                 | ".net" -> Some MSbuildHost.MSBuildExe |> Promise.lift
                 | ".net core" -> Some MSbuildHost.DotnetCli |> Promise.lift
                 | "ask at first use" -> pickMSbuildHostType ()
--- a/src/Components/MSBuild.fs
+++ b/src/Components/MSBuild.fs
@@ -90,7 +90,7 @@ module MSBuild =
                 return ho
             }
 
-    let invokeMSBuildPromise project hostPreference target =
+    let invokeMSBuild project target hostPreference =
         let autoshow =
             let cfg = vscode.workspace.getConfiguration()
             cfg.get ("FSharp.msbuildAutoshow", true)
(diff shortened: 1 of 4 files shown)


enricosada:
added configuration 

```json
        "FSharp.msbuildHost": {
          "type": "string",
          "default": ".net",
          "description": "MSBuild host",
          "enum": [
            ".net",
            ".net core",
            "ask at first use"
          ]
        },
```

no `auto` atm, because is not possibile to infer yet. default is `.net` (backward compatibile with current).

Also how picker (for `ask at first use`) is like

image

src/Components/MSBuild.fs
```diff
@@ -14,20 +14,111 @@ module MSBuild =
...
+            )
+
+    let getMSbuildHostType () =
+        match msbuildHostType with
+        | Some h -> Some h |> Promise.lift
+        | None ->
+            logger.Debug "No MSBuild host selected yet"
+            let cfg = vscode.workspace.getConfiguration()
+            let p =
+                match cfg.get ("FSharp.msbuildHost", "auto") with
```
Krzysztof-Cieslak: Default value should be `.net`

enricosada:
- reload after config change, because why not (trying to play a bit with vscode atm).
- fix to use `dotnet` in PATH also on windows (important for me, and usually the best default)
- msbuild command can now invoked programmatically (with options) and will return a promise (so can continue after is done)

enricosada:
@Krzysztof-Cieslak  is ok for me atm (for auto restore)
default is `.net` so is like before for pratically everyone, i'll add autodetect later, when fsac give me the info

Issue #445, opened 2017-06-08
intellisense doesnt work referencing a c# csproj


NinoFloris:
Problem reading assembly 'ReferencedCsProject/bin/Debug/netstandard1.6/ReferencedCsProject.dll': Build was not evaluated, expected the results to be ready after 'Eval'.

Might this be because the referenced assembly is from a C# project?

enricosada:
@NinoFloris can you share a repro? 
If is F# + C# is not supported atm, see https://github.com/fsharp/FsAutoComplete/issues/171

NinoFloris:
Sorry for the late reply, life got in the way. 

Yep it's mixed F# and C#  :/
Thanks a lot for your work in this space. I see we just have to wait a little for https://github.com/fsharp/FsAutoComplete/pull/173 to get mainlined right?

enricosada:
@NinoFloris yes

Krzysztof-Cieslak:
Fixed in latest FSAC version.

NinoFloris:
New ionide coming soon with it? Thanks for the work, very useful fix.

Krzysztof-Cieslak:
It was just released

enricosada:
@NinoFloris https://github.com/ionide/ionide-vscode-fsharp/wiki/Version-2.27.12#fix-csproj-p2p require the c# project to be built and assembly generated. see notes

PR #467, opened 2017-07-14
Enhance `F# language service` panel logging

enricosada:
Enhance `F# language service` panel logging

**bump FSAC, to use new detailed error messages**

Previous behaviour:

- FSAC log was really verbose, useful errors buried in debug of request/response chatting
- by default, no info to user, who need to enable that verbose logging (and usually doesnt know how and worse that the option exists), for see errors (like failed loading of projects)

New behaviour:

- FSAC request/response log is now DEBUG
- default  (`INFO` like before) show useful info (not yet) and error
- new FSAC etailed errors are pretty printed
- hot update the current log level when configuration change, without need to restart code

Some examples:

### project not restored

image

### some error running msbuild

image

### hot reload config

enh_log

### DIFF
--- a/paket.lock
+++ b/paket.lock
@@ -5,10 +5,10 @@ NUGET
     Octokit (0.24)
 GIT
   remote: https://github.com/fsharp/FsAutoComplete.git
-     (9b91ed87d298cd32551b5cbfa79fef0f1169f026)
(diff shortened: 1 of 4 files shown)


enricosada:
@Krzysztof-Cieslak  FSAC bump will fix #445 too

Krzysztof-Cieslak:
👍

PR #475, opened 2017-07-21
Don't specify full path to vscode.

fsoikin:
The latest (as of July 20th) VSCode Windows installer installs to "Program Files", not to "Program Files (x86)", so the `platformTool "code"` call fails now. At the same time, the modern installer does put `code.cmd` on path.


Krzysztof-Cieslak:
Thanks

PR #494, opened 2017-08-14
Show the project view only when there are projects to see

vbfox:
Don't always show the "Project Explorer" view, only show it when there are some project present (No need to take place on the screen when users aren't doing F# in the workspace).

* Remove a log I forgot in previous PR (oops)
* Rename "Project Explorer" to "F# Project Explorer" (To better distinguish it between the other view when multiple plugins contribute one) *I'm not especially attached to that, but it feel a little better*

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -241,7 +241,8 @@
       "explorer": [
         {
           "id": "ionide.projectExplorer",
-          "name": "Project Explorer"
+          "name": "F# Project Explorer",
+          "when": "fsharp.project.any"
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


Krzysztof-Cieslak:
Huh... what does `setContext` command do? I can't find any documentation about it.

vbfox:
I don't know if it's documented except in the vscode extension samples (the vim one does it for example)

It set a context value at the global level (Like the file/folder/project type is at the tree view node level)

I implemented it as any type but I think the command can only be used with bools (Internally contexts can be any type)

There is an issue to create a true API instead of relying on a command: https://github.com/Microsoft/vscode/issues/10471

vbfox:
https://github.com/Microsoft/vscode/issues/10471#issuecomment-240031506

> vscode.commands.executeCommand('setContext', 'inCordovaProject', true); works today. It is a "sort-of-hidden" API that we introduced for vim extensions in particular. We did not want to make it a top level API concept until we found more use-cases, so having found more use-cases, we are now considering making it a top level API concept, as @jrieken suggests, perhaps via a createContextKey method. We will keep this issue around to track this work.

PR #510, opened 2017-08-26
"LineLens"

vbfox:
Mostly opening the PR to get some feedback :) *The name is not very original ("On the line CodeLens" -> "LineLens") i'm open to suggestion if it's not good*

The concept is to show type signatures at the end of the line like the vscode debugger is showing variable values (See https://github.com/Microsoft/vscode/pull/16129)

linelens1

How to test:

For now it's enabled by default when code lenses are disabled :

```json
{
    "editor.codeLens": false,
    "FSharp.lineLens.enabled": "replaceCodeLens",
    "FSharp.lineLens.color": "theme(editorCodeLens.foreground)",
    "FSharp.lineLens.prefix": "  // ",
    "FSharp.lineLens.refreshAfter": 1000,
    "workbench.colorCustomizations": {
        "editorCodeLens.foreground": "#4C4B4B"
    }
}
```

A few things I need to do / know for this PR :
* [x] Dispose the decoration type when disabled instead of the current hack
* [x] Cleanup line in CodeLens.fs
* [x] Remove the hack to disable `npm update` during build. I don't think update should run during build but I can open another PR/issue for that.
* [x] Wait for https://github.com/ionide/ionide-vscode-helpers/pull/11 & update paket lock
* [x] Disable them by default until we're sure it's working well ?
* [x] Determine if I need to call parse on FSAC or can wait for a parse somehow

Regarding the last one it's mostly a question for @Krzysztof-Cieslak I think, i'm currently calling `LanguageService.parse` manually but isn't it done automatically somehow when file change ? (CodeLens seem to get the change directly) I added this step as otherwise when a file is opened for the first time `LanguageService.declarations` return an error that the file is not yet parsed.

Is it the correct way or will I trigger extra parsing of the opened file ? What I would like is an event from FSAC telling me that the file has changed and been parsed but I didn't find one.

### DIFF
--- a/src/Components/LineLens.fs
+++ b/src/Components/LineLens.fs
@@ -1,0 +1,329 @@
+module Ionide.VSCode.FSharp.LineLens
+
+open System
+open System.Collections.Generic
+open Fable.Core
+open Fable.Import
+open Fable.Import.vscode
+open Fable.Import.Node
+open Fable.Core.JsInterop
+open DTO
+open Ionide.VSCode.Helpers
+
+type Number = float
+
+let private logger = ConsoleAndOutputChannelLogger(Some "LineLens", Level.DEBUG, None, Some Level.DEBUG)
+
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module LineLensConfig =
+    open System.Text.RegularExpressions
+
+    type EnabledMode =
+    | Never
+    | ReplaceCodeLens
+    | Always
+
+    let private parseEnabledMode (s: string) =
+        match s.ToLowerInvariant() with
+        | "never" -> Never
+        | "always" -> Always
+        | "replacecodelens"
+        | _ -> ReplaceCodeLens
+
+    type LineLensConfig = {
+        enabled: EnabledMode
+        codeLensEnabled: bool
+        color: U2<string, ThemeColor>
+        prefix: string
+    }
+
+    let defaultConfig = {
+        enabled = ReplaceCodeLens
+        codeLensEnabled = true
+        color = "editorCodeLens.foreground" |> ThemeColor |> U2.Case2
+        prefix = " //  "
+    }
+
+    let private themeRegex = Regex("\s*theme\((.+)\)\s*")
+
+    let private parseColor (color: string) =
+        let m = themeRegex.Match(color)
+        if m.Success then
+            U2.Case2 (ThemeColor (m.Groups.[1].Value))
+        else
+            U2.Case1 color
+
+    let getConfig () =
+        let cfg = workspace.getConfiguration()
+        let fsharpCodeLensConfig = cfg.get("[fsharp]", JsObject.empty).tryGet<bool>("editor.codeLens")
+        let globalCodeLensConfig = cfg.get("editor.codeLens", defaultConfig.codeLensEnabled)
+        let codeLensEnabled = defaultArg fsharpCodeLensConfig globalCodeLensConfig
+        {
+            enabled = cfg.get("FSharp.lineLens.enabled", "replacecodelens") |> parseEnabledMode
+            codeLensEnabled = codeLensEnabled
+            prefix = cfg.get("FSharp.lineLens.prefix", defaultConfig.prefix)
+            color = cfg.get("FSharp.lineLens.color", "theme(editorCodeLens.foreground)") |> parseColor
+        }
+
+    let isEnabled conf =
+        match conf.enabled, conf.codeLensEnabled with
+        | Always, _ -> true
+        | ReplaceCodeLens, false -> true
+        | _ -> false
+
+module Documents =
+    type Cached = {
+        /// vscode document version that was parsed
+        version: Number
+        /// Decorations
+        decorations: ResizeArray<DecorationOptions>
+        /// Text editors where the decorations are shown
+        textEditors: ResizeArray<TextEditor>
+    }
+
+    type DocumentInfo = {
+        /// Full path of the document
+        fileName: string
+        /// Current decoration cache
+        cache: Cached option
+    }
+
+    type Documents = Dictionary<string, DocumentInfo>
+
+    let inline create () = Documents()
+
+    let inline tryGet fileName (documents: Documents) = documents.TryGet fileName
+
+    let inline getOrAdd fileName (documents: Documents) =
+        match tryGet fileName documents with
+        | Some x -> x
+        | None ->
+            let value = { fileName = fileName; cache = None }
+            documents.Add(fileName, value)
+            value
+
+    let inline set fileName value (documents: Documents) = documents.Add(fileName, value)
+
+    let update info (decorations: ResizeArray<DecorationOptions>) version (documents: Documents) =
+        let updated =
+            { info with
+                cache =
+                    Some {
+                        version = version
+                        decorations = decorations
+                        textEditors = ResizeArray() } }
+        documents |> set info.fileName updated
+        updated
+
+    let inline tryGetCached fileName (documents: Documents) =
+        documents
+        |> tryGet fileName
+        |> Option.bind(fun info -> info.cache |> Option.map(fun c -> info, c))
+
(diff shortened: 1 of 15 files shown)


src/Components/LineLens.fs
```diff
@@ -0,0 +1,376 @@
...
+        |> Array.groupBy(fun r -> r.StartLine)
+        |> Array.choose (fun (_, ranges) -> if ranges.Length = 1 then Some (ranges.[0]) else None)
+
+    let private updateDecorationsForEditor (textEditor : TextEditor) state = promise {
+        let fileName = textEditor.document.fileName
+        let version = textEditor.document.version
+        let text = textEditor.document.getText()
+
+        // TODO: Other things parse doc, how to wait for that ?
+        let! _ = LanguageService.parse fileName text version
```
enricosada: You are already waiting until the parse request, with the `let!`
Atm the parse is cached FSAC side, so should be fast, but there isnt yet an event in ionide side triggered after a document is parsed (or when fsac has sent a `parse` response)

src/Components/LineLens.fs
```diff
@@ -0,0 +1,376 @@
...
+
+    let removeOnChangedLines (changes: ResizeArray<TextDocumentContentChangeEvent>) (textEditor: TextEditor) state =
+        let doc = textEditor.document
+        match state.documents |> Documents.tryGet doc.fileName with
+        | None -> ()
+        | Some info ->
+            match info.cache with
+            | None -> ()
+            | Some cache ->
+                let toRemove = new HashSet<Number>()
```
enricosada: extract this as function and do two `Option.iter` ?

enricosada:
Love the idea, awesome @vbfox !

Because doenst move the layout adding lines like codelens (this just add characters to existing lines), can be really nice to trigger with a keyboard shortcut/command.

src/Components/LineLens.fs
vbfox: Ok the cache was my main question :) I didn't want to end up with the same version being parsed twice juste because LineLens are enabled.

Krzysztof-Cieslak: > Atm the parse is cached FSAC side...
This is true partially... as the response will be cached if the `version` was already parsed. 

I guess I would follow same pattern as Linter and CodeLens is using - define refresh event in module (https://github.com/ionide/ionide-vscode-fsharp/blob/master/src/Components/Linter.fs#L16) which is triggered when file is parsed (https://github.com/ionide/ionide-vscode-fsharp/blob/master/src/Components/Errors.fs#L41)

Krzysztof-Cieslak: If you will update only when this event is called, you will get automatic handling for file open. You'll also be able to drop all the code you have for controlling when to refresh LineLenses

vbfox: Ah thanks, that's exactly what I needed when I started. I wondered where CodeLens got it's updates.

It'll simplify some things & complexify others as I don't want them to appear on lines the user is typing (Decorations are applied asynchronously and can produce bugs where the user type after them)

At least it'll ensure no double-parse

Krzysztof-Cieslak: CodeLenses are bit more complex as they're also refreshed internally by VSCode, not only on the refresh I call (which causes some problems). But anyway... I'd go with event triggered after parse is completed, at least it will work same as other modules

vbfox: Oh i'll do the change it's better all around. I'm just not sure that it simplifies the logic a lot but it'll work better for sure.

The bit about code lenses auto updating was why I didn't dig too much into the event, I assumed that it was linked to the refresh by vscode not by somewhere else in the extension.

vbfox:
@enricosada On the trigger with a keyboard shortcut I wondered about that myself.

Maybe a command that would also be present on the top right of the editor (Along with open change & split) if doable. Anyway it'll be for latter for now i'd like to lend it as a CodeLens replacement :)

vbfox:
Ok new take, mostly a rewrite of the core at this point. Require https://github.com/ionide/ionide-vscode-helpers/pull/12

A few points; 
* It uses the parse event from Error.fs but I Inverted the logic of the event for LineLenses compared to CodeLens. Instead of Error.fs calling into LineLens, it expose a vscode event and LineLens subscribe to it.
* I removed all handling of typing in the current line, it was buggy and I discovered that most of my problems were due to a bad handling of the fact that when the TextDocument version evolve it invalidate all ranges, and anything computed on the previous version should be dropped.
   The fix can be seen in how `updateDecorationsForDocument` check the document version essentially after each async operation and bail out if it changed.
   It also massively simplify the code, so not a bad thing 😉 
* I now cache text editors where I set the decorations to never set them twice for the same editor/version combination 
* When the editors are split it can mean a new TextEditor instance (Re-opening an already opened document a second time) , it's now handled and the decorations are correctly set
* In the same vein the lenses are now correctly set when a document is in more than 1 panes at a time and being edited
* There is now "time before update" anymore it follow the 1s settimeout of Error.fs
* I disabled LineLens by default (Instead of the previous `replaceCodeLens` mode) so that it can be merged and tested by a few ~~guinea pigs~~ **nice users** before a more general release.

Oh I activated CodeLens to see if I didn't break them and it seem I forgot how annoying hey are, let's say that I prefer my little hack :)

vbfox:
Ah also 2 small notes / gotchas:

* `replaceCodeLens` doesn't replace CodeLens if it's globally enabled but disabled only for FSharp language. I tried but can't seem to get the setting. Maybe I should ask vscode team
*  sometimes LineLenses produce `TextEditor disposed` errors. It's not really a problem and caused by the fact that "SetDecoration" is actually asynchronous and that the editor can have disappeared between when we call it and when it really execute. Does not affect any functionality.
   2017-08-30 23_07_28-settings json ionide-fs visual studio code

vbfox:
Turns out I found how to get language specific settings. It's a little manual / weird but works :)

vasily-kirichenko:
I like the IDEA approach more.

vbfox:
@vasily-kirichenko Any more constructive comments ? I don't use IDEA so don't know about their approach to type signature display or how it differ from any of the options available today in ionide...

vasily-kirichenko:
@vbfox Sorry that I was unclear. IDEA shows inferred types and parameter names like this:

image

Kotlin cannot infer parameter types, so the hint makes sense for return types and local vars only.

vbfox:
Ah, very interesting and actually totally doable UI-wise via vscode API.

Personally I don't think i'll like it as one of the reason I dislike codelens is that it mess with my view of the spacing of the code (vertically, but here it would change it horizontally).

But it might be fun to implement, also showing types only when not fully specified seem good, maybe I should do that for LineLens too.

`let x: float = 0. // float` seem like a pretty useless annotation

vasily-kirichenko:
I agree that the IDEA hints add a lot of noise and changes the structure of code. I ended up assigning a keyboard shortcut to turn the feature on and off and it works pretty well: I turn it off while writing new code and turn it on while exploring existing codebase. However, it has some setting:

image

I find it especially useful not as type hints, but as parameter name ones (which would work nicely for tupled functions and methods only in F# though).

Anyway, such a feature would be a great addition to Ionide, especially is provided with decent set of user settings.

enricosada:
Also IDEA seems nice, for some scenario.

Having the full signature at end `// string -> int` seems faster to read for me, for small functions/values. And never mess with scroll or code position (big bonus point)

But IDEA version can help when there are multiple parameters, like `(fun x y z -> x)` or if multiple signature are needed in same line, so both have points. 
About parameters, depends also if these have parens or not, because without parans can be a bit messy

I use that the same as @vasily-kirichenko , to inspect api, so is not a big drawback the moving if i can quick enable/disable. But i think i'll leave @vbfox version always enabled.

Krzysztof-Cieslak:
@vasily-kirichenko: Good suggestion, I'll create feature-request issue when I'll be back in place with decent internet.

@vbfox: Is it ready to merge? If so, do you want to rebase it to create some nice history, or should I squash everything?

vbfox:
@Krzysztof-Cieslak squashed & rebased :)

PR #513, opened 2017-08-27
Fix #512. Fixes error when spawning dotnet path

### LINKED ISSUE #512: Expecto runner can't run dotnet on Windows or macOS
I get  the following error in the Expecto output when running the Expecto Run task: `"/usr/local/share/dotnet/dotnet" ENOENT `when running the Expecto Run task in macOS. I get a similar error in Windows (with a different path). I think this is  related to the string formatting  that was done in 5d351210a7ca7f1f7771693457397eef41463753. I'm guessing this was to avoid problems with spaces? If that's the case it looks like the Fable Compiler may already have this covered: https://github.com/fable-compiler/Fable/issues/249. 

Reversing the changes from that commit fixes the issue for me on Windows.

johngalambos:
You may have something else in mind, but maybe this will help. I just rolled back https://github.com/ionide/ionide-vscode-fsharp/commit/5d351210a7ca7f1f7771693457397eef41463753.

### DIFF
--- a/src/Core/Project.fs
+++ b/src/Core/Project.fs
@@ -200,7 +200,7 @@ module Project =
     let private execWithDotnetWithShell cmd =
         promise {
             let! dotnet = Environment.dotnet
-            return Process.spawnWithShell (sprintf "\"%s\"" dotnet)  "" cmd
+            return Process.spawnWithShell (sprintf "\"%s\"" dotnet "" cmd
         }
 
     let private execWithShell exe cmd =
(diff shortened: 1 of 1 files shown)


Krzysztof-Cieslak:
Can you check if `dotnetExecuteWithShell` works with this fix?

You can check it by running `Run` from right-click context menu in project explorer on any executable .Net Core project. (for example test project using Expecto)

johngalambos:
Whoops--sorry. I looks like those functions work differently and `execWithDotnetWithShell` won't work properly without the `sprintf` call. I've restored it. It doesn't result in the same error as when running `execWithDotnet`.

Krzysztof-Cieslak:
Great, thanks!

PR #517, opened 2017-08-28
Goto #load

vbfox:
Implement "Go to Definition" for `#load`

gotoload

Ideally this should be in FCS and maybe one day i'll dig into it. But I also want to get Completion working for path in `#r`/`#I`/`#load` so I need to have a parser in the extension anyway.

### DIFF
--- a/src/Components/Definition.fs
+++ b/src/Components/Definition.fs
@@ -30,6 +30,7 @@ module Definition =
         }
 
     module FromLoad =
+        /// Remove escaping from standard (non verbatim & non triple quotes) string
         let private unescapeStandardString (s: string) =
             let mutable result = ""
             let mutable escaped = false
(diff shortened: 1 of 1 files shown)


enricosada:
> Ideally this should be in FCS and maybe one day i'll dig into it

@vbfox yes, FCS is the best place. but with really minor changes, this can be moved inside https://github.com/fsharp/FsAutoComplete at least, the language service. Will help a lot maintance, and is a lot easier to test.

So for ionide, the `LanguageService.findDeclaration` call will be transparent, and not a special case.

As you can see from some test output, pratically that return the file + line + col (last two can be ignored in your case)

The code of `findDeclaration` is https://github.com/fsharp/FsAutoComplete/blob/45bf4a7255f8856b0164f722a82a17108ae64981/src/FsAutoComplete.Core/Commands.fs#L261-L262

Bonus point: will help vim/emacs too, and is easier to maintain

Krzysztof-Cieslak:
While I agree that it should preferably be in FSAC... this is really cool feature, so I'll merge it :)

enricosada:
@vbfox if you want to port it to FSAC, i can help guide you in the first PR

After that, become transparent for ionide, and is just some line deleted

dsyme:
nice feature :)

PR #553, opened 2017-09-21
add `Debug` and `Set at startup project` commands

enricosada:
to F# project explorer, if project is a console app

- `Debug` will run the debugger of selected project. no launch.json needed. this just fix a bug in the previous @Krzysztof-Cieslak implementation

lookma_nolaunchjson

NOTE the gif is not updated, the `Debug` is below `Run` in the context menu

- `Set at startup project` update launch.json or create it if not exists

setasstartup

**NOTE** for .NET full (`"type": "clr") the debugging on windows works like before (for old or new sdk). but project must be configured as x64 and portable pdb

```xml
<PlatformTarget>x64</PlatformTarget>
<DebugType>portable</DebugType>
```

### DIFF
--- a/paket.lock
+++ b/paket.lock
@@ -743,10 +743,10 @@ GIT
      (826aa0690b4a1fb536485cc2806fa73b451fcee7)
 GITHUB
   remote: ionide/ionide-vscode-helpers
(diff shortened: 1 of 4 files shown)


Krzysztof-Cieslak:
`Set at startup project` should be called `Generate run and debug config` - I have something else in mind regarding setting startup project. Tasks for restore & build should be generate

PR #533, opened 2017-09-13
update for fsac single console

enricosada:
ref https://github.com/fsharp/FsAutoComplete/pull/212

now fsac has one single console app `fsautocomplete.exe`

to enable old suave mode:

`--mode http --port <PORT>`

This pr also add the usage of `--hostPID` argument 
ref https://github.com/fsharp/FsAutoComplete/pull/190
this enable FSAC to track vscode process, and quit when that process is
terminated (so will avoid zombies if vscode is not terminated gracefully)

PR #556, opened 2017-09-22
Revert "update for fsac single console"

Krzysztof-Cieslak:
Reverts ionide/ionide-vscode-fsharp#533


Krzysztof-Cieslak:
CC: @enricosada - parse request was failing - always returned null

Issue #574, opened 2017-10-05
Add support for MacOS TouchBar commands


Krzysztof-Cieslak:
Using new API - https://code.visualstudio.com/updates/v1_17#_touch-bar-support

Probably what we want to put there:
* Start default project with debugger (`fsharp.debugDefaultProject`)
* Start default project without debugger (`fsharp.runDefaultProject`)
* Add new project (`fsharp.NewProject`)
* Start FSI (`fsi.Start`)
*  Send file to FSI (`fsi.SendFile`)

Also, possibly others depending on current context. Suggestions welcomed.

Krzysztof-Cieslak:
Fixed by #575

PR #575, opened 2017-10-05
wip touchbar commands

baronfel:
All icons are activated when in the fsharp context only.  Addresses #574.

img_20171005_150030

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -672,6 +672,33 @@
           "group": "1_modification",
           "when": "editorTextFocus && editorLangId == 'fsharp'"
         }
+      ],
+      "touchBar": [
+        {
+          "command": "fsharp.debugDefaultProject",
+          "group": "fsharp.project",
+          "when": "editorLangId == 'fsharp'"
+        },
+        {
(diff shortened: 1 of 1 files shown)


baronfel:
The spacing is unfortunate, but I can't seem to remove the >/< markers from the bar.

Krzysztof-Cieslak:
Run, Debug, and New should probably be visible always when user have F# project in current workspace - you can try same condition we use for enabling project explorer - https://github.com/ionide/ionide-vscode-fsharp/blob/master/release/package.json#L340
Start FSI - when there is project OR current file is F# file
Send to FSI - as you've done, when F# file is open (we may later move to showing it only if `.fsx` file is open [but that needs some upstream changes] )

baronfel:
That's better for conditions, but I'm still getting slivers for icons instead of the images and names.  I'll keep banging away at it.  It's nice to see the number of buttons change as I move around into different contexts, though.

baronfel:
ok, it seems making them all a single group makes things behave more.  This is ok for me now.  We can make it better with the filtering and touchbar-specific icon PRs later.

PR #576, opened 2017-10-05
Better QuickInfo

vbfox:
* Hide QuickInfo when there is no signature to display (Not F# or not on something with a signature)
* Hide QuickInfo when the document is closed (Important for the last one)
* Update QuickInfo when a document is parse (They show when the F# project finishes loading and update when the type of the thing under the cursor is known/changed)

Also:
* Fix 30min max on watch mode


Krzysztof-Cieslak:
Thanks!

PR #581, opened 2017-10-07
Improve handling of xml documentation tags

### LINKED ISSUE #572: Special documentation tags not being stripped from tooltips
I'm on Ionide-fsharp 3.3.4 and VSCode 1.16.1

image

VS2017 for comparison:

image

ovatsus:
Fixes #572 

- [x] Treat the content of `<see>` as code in the same way as `<c>`
- [x] Strip out `<remarks>` and leave the inner content
- [x] Replace `<param name="N">content</param>` with `N - content` (and the same for typeparam)
- [x] Unescape `&quot;`, `&apos;`, and `&amp;`
- [x] Handle tags that span multiple lines correctly

image

### DIFF
--- a/src/Core/Utils.fs
+++ b/src/Core/Utils.fs
@@ -165,22 +165,35 @@ module Array =
 
 module Markdown =
     open System.Text.RegularExpressions
+
+    // reference: https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/xml-documentation
+    // TODO: support <exception> and <seealso>
     let private replacePatterns =
         let r pat = Regex(pat, RegexOptions.Compiled ||| RegexOptions.IgnoreCase)
-        [ r"<c>(((?!<c>)(?!<\/c>).)*)<\/c>", sprintf "`%s`" ]
+        let code = Array.item 0 >> sprintf "`%s`"
+        let returns = Array.item 0 >> sprintf "returns %s"
+        let param (s: string[]) = sprintf "`%s` - %s"(s.[0].Substring(1, s.[0].Length - 2)) s.[1]
+        [ r"<c>((?:(?!<c>)(?!<\/c>).)*)<\/c>", code
+          r"""<see\s+cref=(?:'[^']*'|"[^"]*")>((?:(?!<\/see>).)*)<\/see>""", code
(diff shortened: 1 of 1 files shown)


ovatsus:
Ready for review now

src/Core/Utils.fs
```diff
@@ -194,18 +212,15 @@ module Markdown =
...
+        |> String.replace "&apos;" "'"
+        |> String.replace "&amp;" "&"
+        |> replaceXml
+        |> String.split [|'\n'|]
+        |> Array.filter (not << String.IsNullOrWhiteSpace)
+        |> Array.map String.trim
         |> String.concat "\n\n"
-        |> String.trim
-        |> (fun v -> MarkdownString (v))
+        |> (fun v -> MarkdownString v)
```
Krzysztof-Cieslak: This can be replaced with just `|> MarkdownString`

src/Core/Utils.fs
```diff
@@ -194,18 +212,15 @@ module Markdown =
...
-        |> (fun v -> v.Split '\n')
-        |> Array.filter(String.IsNullOrWhiteSpace>>not)
-        |> Array.mapi (fun i line ->
-            let v =
-                if i = 0 && not (String.IsNullOrWhiteSpace line)
-                then "\n" + line.Trim()
-                else line.Trim()
-            replaceXml v
-            )
+        |> String.replace "&quot;" "\""
```
Krzysztof-Cieslak: It may be good idea to put those replacements into separate list, and fold over it - it will make it bit easier to add next replacements in the future if we're still missing something

src/Core/Utils.fs
ovatsus: I tried that, but the compiler complains with `This expression was expected to have type 'string' but here has type 'unit'`. Maybe because the parameter to MarkdownString is optional?

src/Core/Utils.fs
ovatsus: done

src/Core/Utils.fs
ovatsus: Yep, I confirmed it, that syntactic sugar doesn't work when object constructors have an optional parameter

Krzysztof-Cieslak:
Thanks a lot!

PR #585, opened 2017-10-08
Fixes "Full name" emphasis markdown in Tooltip

sasmithjr:
I noticed an error in rendering the markdown for the "Full name" block in the Tooltip. If an identifier contained underscores, it would only apply emphasis to the characters between the final 2 underscores. This changes how the markdown is applied in two ways:

1. Uses MarkdownString to escape the `fullName` text
2. Replaces the `_` with `*` for applying emphasis, described [here](http://spec.commonmark.org/0.28/#emphasis-and-strong-emphasis)

I attempted to use `_`, but it doesn't work even when appropriately escaping any characters in `fullName`; it ended up with all `_` being shown, including the leading and trailing ones. Because `*` isn't an allowable character in identifiers (per [Section 3.4](http://fsharp.org/specs/language-spec/4.0/FSharpSpec-4.0-latest.pdf)), I don't think this issue will occur again.

Old:
name underscore error
name underscore error2

Fixed:
name underscore fix
name underscore fix2

Edit: Issue found on macOS 10.13, VSCode 1.17, Ionide-fsharp 3.5.1

### DIFF
--- a/src/Components/Tooltip.fs
+++ b/src/Components/Tooltip.fs
@@ -35,7 +35,7 @@ module Tooltip =
                         match lines |> Array.splitAt (lines.Length - 1) with
                         | (h, [| StartsWith "Full name:" fullName |]) ->
                             [| yield fsharpBlock h
-                               yield U3.Case2 ("_" + fullName + "_") |]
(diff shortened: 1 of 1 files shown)


Krzysztof-Cieslak:
LGTM

PR #604, opened 2017-10-20
Delete the current line where unusedImport's quick fix is apply

MangelMaxime:
This PR try to delete the line where the fix `unused open` for is applied.

I am not really sure of the implementation because, I don't force the line where the fix is applied any more. Instead I really on the behaviour of VSCode, which seems to force us to select line where we apply the fix.

If needed, I can take some time to search if we can ensure the line where we apply the command.

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -88,7 +88,7 @@
     <Exec Command='$(PaketCommand) restore --project "$(MSBuildProjectFullPath)"' Condition=" '$(PaketRestoreRequired)' == 'true' " ContinueOnError="false" />
 
     <!-- This shouldn't actually happen, but just to be sure. -->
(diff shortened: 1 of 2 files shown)


Krzysztof-Cieslak:
I'd imagine that we need to pass any argument to `editor.action.deleteLines` . How Code knows which line to remove otherwise?

Krzysztof-Cieslak:
Going through VSCode code base it looks like lines that are deleted by `editor.action.deleteLines` are calculated based on the current selection. So I'd try selecting the open statement first, and then call `editor.action.deleteLines`

https://github.com/Microsoft/vscode/blob/2ddfa2ec392823e4cd5ccdfe608d38f6c4f7fff3/src/vs/editor/contrib/linesOperations/common/linesOperations.ts#L219-L298

MangelMaxime:
Indeed, `editor.action.deleteLines` is based on the selection and as to open the quick fix we need to be on the line this is working.

I agree, we should enforce the selection I will update the PR with it.

MangelMaxime:
Ok this time the implementation should be better. I calculated the range to remove in the editor.

Krzysztof-Cieslak:
Perfect, thanks!

PR #616, opened 2017-10-27
support .net core FSAC too

### LINKED ISSUE #78: run ionide with .NET Core
Following twitter discussion, what's missing to run ionide with .NET Core?

**If it's in the roadmap to use .netcore, otherwise just close this issue, np**
I understand atm .net core seem in flux, but works, netcore is going rtm end of month.
Tooling instead it's not going to improve a lot soon, so is project.json for some time until c# start use msbuild instead of project.json, and we need time after that to use it.
But support need time, so i understand if people want to wait.
- [x] FCS ref https://github.com/fsharp/FSharp.Compiler.Service/issues/465

from `paket.dependencies`
- [ ] `ionide/ionide-vscode-helpers`
- [ ] `ionide/FSharpFormatting.CLI`
- [ ] `ionide/FsAutoComplete`
- [ ] `fsprojects/Forge` (or can be excluded for now)
- [ ] `ionide-fsgrammar`
- [ ] ... ?

@Krzysztof-Cieslak  i'd like to help to bootstrap the .net core support of F# projects, but i dont know what prioritize atm after suave.

enricosada:
**NOTE this require (atm) that vscode is started with `msbuild` in PATH**

fix #78

This add fsac running on .net core. that mean atm:

- run as FDD .net core 2.0 app. same binaries, run xplat.
- if fsproj target `netstandard`/`netcoreapp`, .net core sdk is the only requirement (no need for mono)
- new project parser for old fsproj, replace `projectcracker`. Require anyway .net/mono installed, because old sdk proj require .NET Full and .NET full sdk.
- new fsx args references parser for BCL references (like `system`). Require anyway mono installed, because fsx/fsi require .NET Full. But works better
- different way to see what are the .net version installed (on mono too)
- use suave v2 instead of v1

add new setting `FSharp.fsacRuntime`:

- `net` -> the .net version (default, previous behaviour)
- `netcore` -> the .net core version

the FSAC .net core is a FDD .net core console app, started as

```
dotnet release/bin_netcore/fsautocomplete.dll
```

and bundled inside ionide (like other fsac in `release/bin`)

ionide_fsac_netcore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ paket-files
 issueList.md
 release/package-lock\.json
 *.vsix
+/release/bin_netcore/
--- a/build.fsx
+++ b/build.fsx
@@ -64,6 +64,8 @@ let vsceTool = lazy (platformTool "vsce" "vsce.cmd")
 let releaseBin      = "release/bin"
 let fsacBin         = "paket-files/github.com/fsharp/FsAutoComplete/bin/release"
 
+let releaseBinNetcore = releaseBin + "_netcore" 
+let fsacBinNetcore = fsacBin + "_netcore"
 
 // --------------------------------------------------------------------------------------
 // Build the Generator project and run it
@@ -101,6 +103,14 @@ Target "CopyFSAC" (fun _ ->
 
     !! (fsacBin + "/*")
     |> CopyFiles releaseBin
+)
+
+Target "CopyFSACNetcore" (fun _ ->
+    ensureDirectory releaseBinNetcore
+    CleanDir releaseBinNetcore
+
+    !! (fsacBinNetcore + "/*")
+    |> CopyFiles releaseBinNetcore
 )
 
 let releaseForge = "release/bin_forge"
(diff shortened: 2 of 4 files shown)


paket.dependencies
```diff
@@ -7,10 +7,10 @@ nuget Fable.Import.Node
 nuget Fable.Import.Browser
 nuget FSharp.Core
 
-git https://github.com/fsharp/FsAutoComplete.git more_fixes build:"build.cmd LocalRelease", OS: windows
+git https://github.com/enricosada/FsAutoComplete.git netcore build:"build.cmd LocalRelease", OS: windows
```
vasily-kirichenko: 👍

enricosada:
@rneatherway @kjnilsson this is how can be added the .net core version of fsac easy, if you want to add that in vim/emacs too when ready.

- bundle the `bin/release_netcore/` generate by FSAC build (contains `fsautocomplete.dll` console app)
- invoke that with `dotnet`, like `dotnet /path/to/fsautocomplete.dll` and usual args

enricosada:
@Krzysztof-Cieslak for me is ok. can be a beta.
this will add 6MB more to bundle, but is not the default, and will help a lot to beta test it.

PR #643, opened 2017-11-14
workspace load

enricosada:
use new FSAC `workspaceLoad` command, instead of multiple `project` command

command is async, and project loading status (inprogress/loaded/failed) is back
as notification

add setting `FSharp.workspaceLoader` :

- `projects` send to FSAC multiple "project" command (default, old behaviour)
- `workspaceLoad` send to FSAC the "workspaceLoad" command and use notifications

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -975,6 +975,15 @@
 						"ionideSearch"
 					]
 				},
+				"FSharp.workspaceLoader": {
+					"type": "string",
+					"default": "projects",
+					"description": "Choose the FSAC workspace loader",
+					"enum": [
+						"projects",
+						"workspaceLoad"
+					]
+				},
 				"FSharp.workspaceModePeekDeepLevel": {
 					"type": "integer",
 					"default": 2,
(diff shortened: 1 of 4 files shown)


Krzysztof-Cieslak:
> enabled by setting, default is old way

Meh, people won't use it. Make new way default. People are more likely to look into settings when something is not working.

enricosada:
> Make new way default. People are more likely to look into settings when something is not working.

Yes, i know, but atm i prefer it as not the default, if possibile, for a bit. after that, is just to change the default of setting

- is not that useful yet unless big dotnet/sdk solutions and that's not the usual scenario
- doesnt yet use FSAC cache, so is like a cold load. and that mean slower loading
- it's the foundation for next improvements (newparser, auto-restore if needed, configuration release, watch changes fsac side, etc) but doesnt add value atm

but i'd like to have it merged, so is easier to test it too (i'll enable in all solutions), and can be used while developing other stuff (netcore, fsac wip who doesnt need ionide changes, etc)

Issue #665, opened 2017-11-29
Type signature not fully rendering in bottom left statusbar


gerardtoconnor:
As mentioned on slack, Type signature not fully rendering in bottom left status bar such that the type/member name renders but its appended signature does not.

If you hover over it, the full type and signature is rendered so specific to how it's rendered in the bar

![capture](https://user-images.githubusercontent.com/5388991/33374334-9dc8d6fc-d4fd-11e7-9abf-c623dcd505c4.PNG)

![capture2](https://user-images.githubusercontent.com/5388991/33374364-c5a26ab2-d4fd-11e7-8ebe-d844bd2f7d02.PNG)

Krzysztof-Cieslak:
I think it should work structured signature info we can get instead of taking first line from tooltips

enricosada:
some hints:

https://github.com/ionide/ionide-vscode-fsharp/blob/5709f05df5a1adba848d15405f32b64a756cc8dd/src/Components/QuickInfo.fs#L34

and

https://github.com/ionide/ionide-vscode-fsharp/blob/5709f05df5a1adba848d15405f32b64a756cc8dd/src/Components/QuickInfo.fs#L51

the `.[0]` is why just the first line of tooltip is shown)

PR #667, opened 2017-11-30
Fix taskbar signature render

gerardtoconnor:
fix https://github.com/ionide/ionide-vscode-fsharp/issues/665

I removed the use of tooltip and replaced with signature as it is a better fit for the one line render in taskbar and does not loose all the extra information from tooltips that are multiline


enricosada:
tested locally! awesome @gerardtoconnor !

PR #674, opened 2017-12-10
DOC: improvements to first time contribution experience

daz10000:
Included are some modest improvements for making the first time contribution experience with ionide-fsharp a little smoother.  Took me a few hours to get a successful build.   I'm going to start working on some of the other tickets now I have a build.   

I did bump the fable.core paket version to 1.3.4 as that was blocking my initial build Option.js was missing.

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -1,8 +1,8 @@
 source https://www.nuget.org/api/v2
 storage:none
 
-clitool dotnet-fable 1.2.4
-nuget Fable.Core 1.2.4
+clitool dotnet-fable 1.3.4
+nuget Fable.Core 1.3.4
 nuget Fable.Import.Node 0.1.5
 nuget Fable.Import.Browser 0.1.2
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,7 @@
 - [.Net Core 2.0][dotnet]
 - [Node.js][nodejs]
 - [Yarn][yarn]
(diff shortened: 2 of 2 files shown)


paket.dependencies
```diff
@@ -1,8 +1,8 @@
 source https://www.nuget.org/api/v2
 storage:none
 
-clitool dotnet-fable 1.2.4
-nuget Fable.Core 1.2.4
+clitool dotnet-fable 1.3.4
+nuget Fable.Core 1.3.4
```
enricosada: @Krzysztof-Cieslak is written in the PR description, but to be sure you see that, there is bumping fable.core (you had an issue in upgrade if i remember right)

Krzysztof-Cieslak: Yes, I do believe 1.3 still breaks something in the runtime - I’m working on another Code extension and I also pinned 1.2 there

daz10000: Yes - my experience was that simply running ./build.sh the first time didn't work, and it was due to having an older version of dotnetfable (1.2.4) that doesn't have Option implemented.  Once I ran ./build.sh Build  it pulled down 1.3.4 and it would build happily.   The odd thing is that I then removed the 1.3.4 package from the nuget cache, did a clean pull and it built and didn't seem to install 1.3.4 but I'm pretty sure it's a necessary dependency, so I would bump the version number unless you have a good reason not to.   Overall was a good startup experience and I'm now working on how to improve error messages when other components aren't installed correctly.

PR #675, opened 2017-12-12
Restore projects sync

forki:
/cc @enricosada @Krzysztof-Cieslak @alfonsogarciacaro

I wasn't able to test it, but what do you gys think?

### DIFF
--- a/src/Components/MSBuild.fs
+++ b/src/Components/MSBuild.fs
@@ -224,9 +224,11 @@ module MSBuild =
         let rec messageLoop() = async {
             let! (path,continuation) = inbox.Receive()
             let host = tryGetRightHost' path
-            let _:JS.Promise<unit> = 
+            do!
                 invokeMSBuild path "Restore" (Some host)
                 |> Promise.bind continuation
+                |> Async.AwaitPromise
+
             return! messageLoop()  
         }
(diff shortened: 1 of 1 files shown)


forki:
I assume we still need to await the promise here. How can we do that?

forki:
I guess AwaitPromise should do the trick

src/Components/MSBuild.fs
```diff
@@ -220,6 +220,20 @@ module MSBuild =
         let host = tryGetRightHost' path
         invokeMSBuild path target (Some host)
 
+    let restoreMailBox = MailboxProcessor.Start(fun inbox->
```
enricosada: MailboxProcessor.Start now work in fable too?

alfonsogarciacaro: Yes, there's [limited support for MailboxProcessor](http://fable.io/docs/compatibility.html#fsharp-core) (check the caveats)

forki:
Is it released?

Krzysztof-Cieslak:
Yeah, it was released in `3.14.5` - https://github.com/ionide/ionide-vscode-fsharp/blob/master/RELEASE_NOTES.md#3145---15122017

PR #684, opened 2017-12-22
Fix run Expecto tests unquoted project file path bug.

### LINKED ISSUE #680: Expecto Build fails when there are spaces in the project path
https://github.com/ionide/ionide-vscode-fsharp/blob/212312f4895de8d26eadee94509bda0eb8115f7d/src/Components/Expecto.fs#L242

Having my project in the following path fails: C:/Source Code/Project

If I switch it to C:/SourceCode/Project then it works fine. 

Output:
MSBUILD : error MSB1008: Only one project can be specified.

This seems to be the case because the project path is not being quoted during the call to msbuild.

gdziadkiewicz:
Fixes #680.

Tested with core and classic .net framework projects with Expecto tests and previously failing path.

If we decide to merge it in this form then it would probably be better to change `MSBuild.invokeMSBuild` as it does the quoting thing on it's own.

### DIFF
--- a/src/Core/Utils.fs
+++ b/src/Core/Utils.fs
@@ -53,7 +53,14 @@ module String =
     let endWith ending (s : string) = s.EndsWith ending
 
     let startWith ending (s : string) = s.StartsWith ending
-    let quote (s: string) = sprintf @"""%s""" s
+    let quote (s : string) =
+        let isQuoted (s : string) = s.StartsWith @"""" && s.EndsWith @"""" 
+        let containsWhitespace = Seq.exists Char.IsWhiteSpace
+        let quote = sprintf @"""%s"""
+        match s with
+        | s when s |> isQuoted |> not && s |> containsWhitespace -> quote s
+        | s -> s
+
 
 [<RequireQualifiedAccess>]
 module Option =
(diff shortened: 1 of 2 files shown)


src/Core/Utils.fs
```diff
@@ -85,6 +85,9 @@ module Utils =
         [<Emit("$0.has($1) ? $0.get($1) : null")>]
         member this.TryGet(key: 'key): 'value option = jsNative
 
+    type DTO.Project with
+        member this.ProjectQuoted = String.quote this.Project
```
enricosada: ihmo we dont really need this.

src/Core/Utils.fs
```diff
@@ -53,7 +53,7 @@ module String =
     let endWith ending (s : string) = s.EndsWith ending
 
     let startWith ending (s : string) = s.StartsWith ending
-
+    let quote (s: string) = sprintf @"""%s""" s
```
enricosada: lovely if you can quote only if need (if string contains a whitespace).

gdziadkiewicz:
@enricosada Thanks for feedback :) I've removed the extension and added the quote only if needed part. How do you like it after the changes?

Krzysztof-Cieslak:
LGTM, Thanks!

PR #688, opened 2017-12-25
Add run script and send to FSI in the top right of the editor

vbfox:
Took me quite a time to PR it :) (Used it as an Advent calendar post https://blog.vbfox.net/2017/12/16/hacking-on-ionide.html )

Send and Run buttons

### DIFF
--- a/release/images/run-dark.svg
+++ b/release/images/run-dark.svg
@@ -1,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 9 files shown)


Krzysztof-Cieslak:
Any reason not to use existing send file to FSI functionality?

vbfox:
For the send file ? It's what's used.

For the run I added it as a separate functionality as I sometimes want to run the script as it will be executed in the end without the current FSI env, connected directly to the console. Also potentially running for a long time without cluttering my FSI session.
In the future I also want to add a way to specify cmdline arguments and memorize them between runs.

To be fair there is no need for any of that to be in ionide or even with a big icon enabled by default, it could be behind a setting or even a separate "Run F# script" vscode extension :) As you wish.

Krzysztof-Cieslak:
My brain kinda stopped for the moment, and I haven't noticed that there are actually 2 commands included. 

It's totally fine, and explanation for having both is reasonable, thanks!

Issue #589, opened 2017-10-09
Add a command to enable diagnostics and prepare info for report the issue


enricosada:
When there are issue, usually FSAC log and env version helps a lot.

Is common for diagnose issue to enable the diagnostic log ( ref https://github.com/ionide/ionide-vscode-fsharp/blob/master/README.md#how-to-get-logs-for-debugging--issue-reporting )

Proposed: 

- add a command `F#: enable diagnostics` or something like that, to enable these props (and `disable` too, if are enabled, to revert back)
- add a command: `F#: get info for diagnostics` who get all info about env (os/mono/msbuild/.netcoresdk) + fsac log.

So that user can code/paste here the info, without multiple steps of Q/A ( an example is https://github.com/ionide/ionide-vscode-fsharp/issues/582 )

MangelMaxime:
I would like to take this issue.

PR #694, opened 2018-01-10
Add diagnostics commands (toggle mode + get infos)

MangelMaxime:
Related #589 

# Diagnostics channel

I added a "Diagnotics channel" to handle [this promise failure](https://github.com/ionide/ionide-vscode-fsharp/compare/master...MangelMaxime:diagnostic_commands?expand=1#diff-0b5f9646e998eb81b17c8fea47891233R56) but I don't really like it. How should I report problems here ?

# F#: Toggle diagnostics

I wasn't able to customize the "F#: Toggle diagnostics" to switch display between:
- F#: enable diagnostics
- F#: disable diagnostics

Any ideas ? According to the docs we can access the "user or workspace settings" by adding `config.` prefix. [When close contexts](https://code.visualstudio.com/docs/getstarted/keybindings#_when-clause-contexts).

But the output channels settings are using string. I tried `config.FSharp.logLanguageServiceRequests == 'both'` without success.

Also, in order for the `FSharp.logLanguageServiceRequests` change to be taken into account user need to reload VSCode. Where for `FSharp.logLanguageServiceRequestsOutputWindowLevel` it's detected and loaded on the fly. [Source code](https://github.com/ionide/ionide-vscode-fsharp/blob/master/src/Core/LanguageService.fs#L61-L65) 
Should we do something about that ?

# F#: Get info for diagnostics

I added the following info for the diagnostics:

- OS
- Architecture type
- VSCode version
- Mono version
- Dotnet version
- MSBuild
- Info about the runtime setting ("net" or "netcore")

I don't know how to get the "FSAC channel output". Seems like we can't read it from the ouput and need to access the info before logging it. Any ideas ?

**Important**: I didn't test the code under OSX or Windows because I don't have access to any of them.

-------------
For the infos generated by "F#: Get info for diagnostics", I added some markdown comments:

```
<!-- Please copy/paste this file content into a Github issue -->
### Problem

<!-- Describe here your problem -->

### Steps to reproduce

<!-- Add here the step to reproduce you problem. Example: -->
<!-- 1. Open an F# file -->
<!-- 2. Ctrl + P > "F# Add Reference" -->
```

------------

Here is the result of copy/pasting the infos from my PC.

### Machine infos

* Operating system: **Linux**
* Arch: **x64**
* Runtime: **.Net**
* Mono version:
```shell
Mono JIT compiler version 5.4.1.6 (tarball Wed Nov  8 20:37:08 UTC 2017)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	LLVM:          supported, not enabled.
	GC:            sgen (concurrent by default)
```
* MSBuild version:
```shell
Microsoft (R) Build Engine version 15.4.0.0 ( Mon Nov 27 18:23:43 UTC 2017) for Mono
Copyright (C) Microsoft Corporation. All rights reserved.
```
<details>
<summary>
    FSAC logs (click to expand)
</summary>
<p>

```shell

```

</p>
</details>


enricosada:
As a feedback the channel is the wrong medium.

As basic feature:  adding that to clipboard directly + a message like `diagnostic info added to clipboard` will be really nice and enough 90% of time.

But instead of a channel, dont just open a page? like the `F# Soluition explorer -> right click on project -> Properties`

channel are for stream of data or interactive. afaik you just want to printout some info to xcopy. we have already two channels, adding more is confusing.

is really easy to implement , see https://github.com/ionide/ionide-vscode-fsharp/blob/23897e3f0aac1e85ca2066137f228c12b0bb9b2b/src/Components/SolutionExplorer.fs#L479-L541

create a `TextDocumentContentProvider`, with a  `/ionide/diagnostic` route, and that's it. will be nice to add more stuff later too maybe

enricosada:
> I don't know how to get the "FSAC channel output". Seems like we can't read it from the ouput and need to access the info before logging it. Any ideas ?

save the lines in memory or file, for that session only. like a `tee`.
is a lot of traffic to record everything everytime.
maybe we can ask user to enable diagnostic, save the line to file and add that too

MangelMaxime:
@enricosada Thanks for your feedback.

I am not sure to understand your first message.

If you are speaking about the new "Diagnostic" channel I created I agree it's not good for it's purpose and I should use the `vscode.previewHtml` commands to output the error. *I will fix it*

Other than that, I am not using the channel to output the diagnostics infos. Here is a GIF to show you the current experience.

diag_demo

I will take a look to add the content to the clipboard instead.

Krzysztof-Cieslak:
I like how this gif looks like - I think it's good enough behavior for most cases, and people can just manually copy the content ;)

I also agree that adding `Diagnostic` output channel is not too good - we already have too many output channels and it's getting confusing. 

For posting content of the FSAC output channel - I'd add command that saves the content of the channel to file, and then people will be able to attach the file to the issue on GitHub.

Krzysztof-Cieslak:
Also... in super long term (not in scope of this PR i guess), we could create command that posts issues to GitHub from editor - auth user with GitHub's OAuth and then create issue with API. User would just need to edit details in generated template, press the button and we would create issue and attach file with FSAC channel content etc.

MangelMaxime:
@Krzysztof-Cieslak So you would prefer users to copy/paste the gets infos + find the FSAC logs file and link it to the issue ?

I was planning to add the FSAC logs at the end of the issue under a collapsible area. Like that everything is done with one copy/paste.

And yes I thought about using GitHub's OAuth too but said myself we could indeed add it later. :)

Krzysztof-Cieslak:
Adding logs to the collapsible area would be good too. I'm just bit afraid of length - FSAC logs are rather verbose and for long sessions, it can be a lot of content.

MangelMaxime:
@Krzysztof-Cieslak So finally, I remove the "collapsible area" because perhaps not everyone want to share it by default. It can contain private code etc.

I added a command: "F#: Get FSAC logs", which copy the logs registered for the session into a file: "$HOME/ionide/FSAC_logs/yyyy-MM-dd HH:mm:ss.log".

I tried to open the containing folder but didn't find a way to do it. So instead, user can decided to open the file or go manually to the folder. See the following gif:

demo_get_fsac_logs

**Question:** If you know how to open the containing folder it's should be better.

----

@enricosada @Krzysztof-Cieslak In order, to capture the FSAC logs of the current session, I write them into a file located in the *temp* directory of the system. Like so, the files are removed by the OS when needed.

I first wanted to log all the logger into the file and the finally decided to logs only the ones coming from `"IONIDE-FSAC"` [source code](https://github.com/ionide/ionide-vscode-fsharp/pull/694/files#diff-9406e7f5b56524c45bff635fa8bd6662R61)

**Question:** With the current implementation, I write all levels of logs should we wired it on `chanMinLevel` ?

enricosada:
Another way is we just store in memory the last 100-200 FSAC messages (the json). So is not too taxing as memory (2-3MB? json payload is not too verbose). So we discard when new message go through.

And the `get diagnostic` command, just read that array in memory. Usually we dont need whole session recorded.

Good part: code is already written, just atm write to the channel if verbose diagnostics is enabled. we can make it append to that array instead

My most common issue in the tracker, is a problem with project loading.
i need log info because contains the FSAC response to loading. but that is also stored in the `projects` too (used by `F# project explorer` )

enricosada:
Also, `get FSAC log` is not the best naming, ppl doesnt known about `FSAC`. just `F#: get language server log` or `F#: get ionide logs` or `ionide: get logs`

MangelMaxime:
Storing 100-200 FSAC messages in memory is good to me too.

And I agree for `F#: Get Ionide logs` naming. This would allow us to add others logs than FSAC if needed one day.

MangelMaxime:
Should I wait after @Krzysztof-Cieslak confirmation ?

Krzysztof-Cieslak:
Let's try with 200 messages, and maybe let's make the amount configurable. 

If we take a look at logging - the one thing I maybe would want to get rid of logs is file content in `parse` request. Having this in log may be problematic from people that want to report issue happening on private projects etc.

MangelMaxime:
Ok, I will update the PR accordingly

MangelMaxime:
Sorry for the delay guys was having a rush period at work.

The PR should be ready to review, I think.

PR #707, opened 2018-02-08
Make CodeOutline 🌟🌟🌟

vbfox:
* Outline only appear on F# files and dynamically hide & show itself
* Settings for outline are dynamic: Change your settings in another vscode windows and it refreshes live !
* Settings for outline are workspace aware: Even with multiple workspaces open the setting for each will be respected in their files.

Ah and also when disabled it doesn't less an empty pane with buttons throwing errors ;)


vbfox:
(Require https://github.com/ionide/ionide-vscode-helpers/pull/18 )

PR #708, opened 2018-02-09
Use rename when replacing PascalCase literals

coolya:
If a literal is already in used and the liter suggests to replace it with
a PascalCase variant use the document rename action instead of simple
text replacement. Because otherwise the code is broken after applying
the linters suggestion.


Krzysztof-Cieslak:
Thanks a lot!

PR #715, opened 2018-02-20
Update Windows Installation Link

ShalokShalom:
I think you might like to update the Ionide page as well - while I dont get this description at all, since Ionide installs FSharp anyway?


ShalokShalom:
Fun thing: Microsoft decides to use http in 2017 on their downloads page?

ShalokShalom:
Then stays the outdated link.

PR #729, opened 2018-03-06
Support spaces in arguments to forge

tlycken:
When adding a file `Foo.fs` to my project through the F# Project Explorer context action "Add file above", I noticed that the file path in the project was set to `Bar` instead of `Foo.cs`. After digging around a little, I realized this is because the project is in a folder like `C:\Stuff\Bar Baz\MyProject`, and so the path was being cut at the first space.

I wanted to test this out locally, but I was unable to start the extension in debug mode (because `C:\Program Files (x86)\Microsoft SDKs\F#\4.0\Framework\v4.0\Microsoft.FSharp.Targets` was not found, which is correct since I have `...\F#\4.1\Framework\...` on my system, but unfortunate).

But I figure a patch that you can test is better than no patch at all :)

### DIFF
--- a/src/Components/Forge.fs
+++ b/src/Components/Forge.fs
@@ -59,22 +59,22 @@ module Forge =
         else
             path
     let moveFileUpPath path =
-        sprintf "move file -n %s -u" path |> spawnForge |> ignore
+        sprintf "move file -n '%s' -u" path |> spawnForge |> ignore
 
     let moveFileDownPath path =
-        sprintf "move file -n %s -d" path |> spawnForge |> ignore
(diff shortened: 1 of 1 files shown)


Krzysztof-Cieslak:
LGTM, thanks a lot!

Krzysztof-Cieslak:
Looks like this introduced the bug reported on the Slack - the quoting that you've introduced shouldn't also include names of the template. I'll reverse for now.

Krzysztof-Cieslak:
Correction: I'll just fix it.

Issue #712, opened 2018-02-12
Suggestion: Make open namespace the first option in the list


rodrigovidal:
HI guys, 

I just want to suggest to make the open namespace the first option rather than using qualified namespace.

![image](https://user-images.githubusercontent.com/388081/36090532-fc5837e4-0fe0-11e8-9c80-ee2fda2c7af7.png)

PR #733, opened 2018-03-08
Swap order of using and open namespace actions

tboby:
This implements #712.

First contribution, is this really all the change requires? It worked locally for me!


Krzysztof-Cieslak:
Looks good to me, thanks a lot! :heart:

Issue #753, opened 2018-03-26
Get Ionide logs failure.


Krzysztof-Cieslak:
https://imgur.com/a/aEdBD

Reported by @landy on Twitter.

CC: @MangelMaxime

MangelMaxime:
I think the problem is coming from https://github.com/ionide/ionide-vscode-fsharp/blob/12c263901638e7f01ceba5e786c911df1a8dd8a9/src/Components/Diagnostics.fs#L211

I was thinking that `Exports.Os.homedir()` was able to find the `HOME` dir on windows, but not sure if all the Windows OS have it by default.

Also, I think we should use `Exports.Os.tmpdir()` instead of `Exports.Os.homedir()` like that the logs will be cleaned by the OS when needed. For now, they are written and never cleaned.

I can send a PR tomorrow, however I am not sure to have access to a Windows computer..

landy:
Let me know if I can help you with something

PR #756, opened 2018-03-27
Related to #753

MangelMaxime:
@landy @Krzysztof-Cieslak I don't have access to a Windows computer so not sure if this will solve the problem.

If the path is still invalid for Windows, then we are probably doing something wrong in this function

https://github.com/ionide/ionide-vscode-fsharp/blob/master/src/Core/Environment.fs#L167-L184

If you have access to a Windows computer can you please test this branch ?

### DIFF
--- a/src/Components/Diagnostics.fs
+++ b/src/Components/Diagnostics.fs
@@ -208,7 +208,7 @@ Error: %s
 
     let getIonideLogs () =
         let writeStream =
-            Exports.Path.join(Exports.Os.homedir(), "ionide", "FSAC_logs")
+            Exports.Path.join(Exports.Os.tmpdir(), "ionide", "FSAC_logs")
             |> Environment.ensureDirectory
             |> fun dir -> Exports.Path.join(dir, DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss.log"))
             |> Exports.Fs.createWriteStream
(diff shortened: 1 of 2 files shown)


landy:
The error is the same but I thing the function ensureDirectory does something wrong.  It receives correct path `C:\Users\landsman\AppData\Local\Temp\ionide\FSAC_logs` but when it computes the tmp value it is set to `\` and then it tries to create directory \C: and then it fails so I think this code https://github.com/ionide/ionide-vscode-fsharp/blob/f1b4063279633cd47a5ca5e93df347f0656ab907/src/Core/Environment.fs#L168-L169 needs to return something else

MangelMaxime:
@landy Thanks for the report can you please try again ?

landy:
now the first subpath value is set to `C:.` Dont know how `.` got in there but I think it doesn't matter. but there is an error when you join this sub path with next segment it creates `C:Users` instead of `C:\Users`

MangelMaxime:
@landy Can you try the latest version please, I needed to include a fix for Unix os...

@Krzysztof-Cieslak I am not really happy with the new `ensureDirectory` implementation (feels hacky) do you have an idea on how to rewrite it ? Should we use an existing js lib ? Keep it like that ?

landy:
works on my computer ;)

landy:
```
let options =  System.StringSplitOptions.RemoveEmptyEntries
let s = "/root/etc"
let ls = s.Split  ([| '/' |], options) |> Array.toList
```

maybe this would do the trick? didn't try it

Issue #770, opened 2018-04-02
Evaluating line ending with comment in FSI doesn't work correctly


ovatsus:
VSCode 1.21.1
Ionide 3.18.2

If you have a line like this: `let x = 2 // comment` and press Alt+Enter either without selection or selecting the whole line, you don't get the proper result in FSI, because the `;;` added at the end is ignored:

![image](https://user-images.githubusercontent.com/738761/38219540-f91d4ccc-36cd-11e8-927c-e5b50629213d.png)

You have to select just the bit until the 2 to get the evaluation to work:

![image](https://user-images.githubusercontent.com/738761/38219550-0231ebb0-36ce-11e8-8d98-de2ccbfb3c23.png)

I'm not 100% sure, but I think this worked before in Ionide.

In real code this isn't much of a problem, but when using Ionide to do a small demo introduction to F# to newcomers it's really annoying

Krzysztof-Cieslak:
Hey, @vbfox. That’s why `;;` were put on new line ;-)

Krzysztof-Cieslak:
Bug was introduced as part of #766

ovatsus:
Ideally you would only put on the new line when there is a comment, but keep on the same line generally

vbfox:
Sorry

PR #772, opened 2018-04-03
Put the ;; on the next line when sending code to FSI when there's a comment

ovatsus:
This will have false positives if there's a // inside a string, but that should be harmless

### DIFF
--- a/src/Components/Fsi.fs
+++ b/src/Components/Fsi.fs
@@ -81,7 +81,7 @@ module Fsi =
             i1 <- i2]
 
     let private send (msg : string) =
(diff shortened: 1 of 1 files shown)


ovatsus:
closes #770

vbfox:
Yes better safe than sorry.

I removed the extra `/n` as it was adding noise but as long as it's removed most of the time I think we're good.

Another case that is always buggy is sending a block comment start `(*` without the end. It's even funnier in VS where the `(*` isn't reflected (The FSI UI never reflect alt-enter) but is still taken into account so anything you send after is ignored without any indication of why.

PR #780, opened 2018-04-08
Fix #773 --no-prompt project directory.

### LINKED ISSUE #773: Ionide creating project under directory called "--no-prompt"
Ionide's project creation seems to be broken. I open VS Code in an empty folder (`Experiment.FSharp`) and run `F#: New Project -> console`, leaving the project directory blank and specifying `Experiment` as the project name. It ends up generating a project hierarchy like this:

```
$ tree -L 3
.
├── build.cmd
├── build.fsx
├── build.sh
├── --no-prompt
│   └── Experiment
│       ├── App.config
│       ├── Experiment.fs
│       ├── Experiment.fsproj
│       └── paket.references
├── packages
│   └── FSharp.Core
│       ├── [Content_Types].xml
│       ├── _rels
│       ├── fsharp.core.4.3.4.nupkg
│       ├── FSharp.Core.nuspec
│       ├── lib
│       └── package
├── paket.dependencies
├── paket.lock
└── paket-files
    └── paket.restore.cached
```

#### Environment

 - Windows 10  Pro 1709
 - VS Code v1.21.1
 - Ionide-fsharp v3.18.2

gdziadkiewicz:
## Problem:

~~Providing empty string in vscode input results with using "--no-prompt" as project directory.~~
When provided with empty project directory string we should pass it to forge call as quoted empty string.

## My solution:

~~I have added empty string check and error message for project directory.~~
Added empty string case to `quotePath`.

## ~~Alternatives:~~

1. ~~Changing Forge behavior as atm Forge needs the project directory - if you don't supply it in args it will prompt for it (even with --no-prompt switch).~~
2. ~~Default, nonempty, project directory?~~

### DIFF
--- a/src/Components/Forge.fs
+++ b/src/Components/Forge.fs
@@ -55,7 +55,7 @@ module Forge =
 
     let private quotePath (path : string) =
         if JS.isDefined path then
-            if path.Contains " " then "\"" + path + "\"" else path
+            if path = "" || path.Contains " " then "\"" + path + "\"" else path
         else
             path
     let moveFileUpPath path =


Krzysztof-Cieslak:
I don’t think it’s good fix - this feature used to work with passing just empty string as dir. 

I think problem is `quotePath dir` part - probably it’s returning empty string when `dir` is empty string and what we need to pass in such case is `””`

gdziadkiewicz:
I have changed the fix.

Krzysztof-Cieslak:
Thanks!

Issue #801, opened 2018-05-09
Feature Request: Sync with active document 


davidtme:
The new F# side project explorer button is great :) Would you be able to add a button/keyboard shortcut that acts like the "Sync with Active Document" in VS2017 and highlights the active file in the project explorer?

![image](https://user-images.githubusercontent.com/1804141/39814686-32082bca-538d-11e8-92a7-2b02ddeccc18.png)

Krzysztof-Cieslak:
Something like that, and code for it is already merged. Unfortunatly due to some bugs on Code side we can't enable feature yet.

PR #808, opened 2018-05-13
Auto-Reveal in project explorer

vbfox:
Implement the same feature as `autoReveal` in the project explorer: the current file is auto selected in the view.

* getParent is implemented in the tree by adding a parent (ref cell) member to all DU cases where it's relevant
* When the tree model changes the tree raise an even with the root that allow the code to generate a `path -> Model` map
* On document change the map is checked and `reveal` is called if found and the feature is enabled.

There is currently an annoying behavior that it change the current activity / expand the explorer view container (Something the main explorer doesn't do). That's why I disabled the feature by default for now. I reported it to vscode : https://github.com/Microsoft/vscode/issues/49779

But even if we can't enable it by default it will allow us to implement https://github.com/ionide/ionide-vscode-fsharp/issues/801 (As it need the same data structures & features)

Require https://github.com/ionide/ionide-vscode-helpers/pull/24
Incorporate https://github.com/ionide/ionide-vscode-fsharp/pull/807 code so the change list might be a little confusing before this one is merged.


MangelMaxime:
Fortunately I found your PR before starting on this feature :D

Well done ^^

PR #833, opened 2018-06-01
Respawn FSAC if it's terminated _not_ by us

### LINKED ISSUE #832: [Resiliency] Should recover when FSAC process dies
There should be some kind of listener on the spawned FSAC process so that if the process exits unexpectedly (for me this happens with a stackoverflow?) a new process would be spawned and initialized with the workspace context.

baronfel:
Fixes #832 by respawning the child process on process exit, but only if we didn't ask for the termination. We have to communicate this via the mutable boolean because node 7.x doesn't give use the signal used to kill the process ([docs here](https://nodejs.org/docs/latest-v7.x/api/child_process.html#child_process_event_exit)).


Krzysztof-Cieslak:
LGTM, thanks.

PR #841, opened 2018-06-08
Updating to FAKE 5 modules (WIP)

nelak:


nelak:
@Krzysztof-Cieslak Is ionide dotnetcore only or does it support mono and .NET?

nelak:
@Krzysztof-Cieslak This is ready for review. 
It now handles user input with `Fake.Core.UserInput` instead of the previous `UserInputHelper`
Octokit is still pinned to version 0.29 since the newer 0.30 broke `octokit.fsx`

MangelMaxime:
I think you are getting a warning when running FAKE with this branch. 

Because you are using the old runner instead of the FAKE tool.

PR #855, opened 2018-07-20
Don't modify markdown strings

### LINKED ISSUE #852: Tooltips including newline when showing XML doc info
### Problem

When XML doc comments are written using hard returns (say at 80 or 120 characters), each line is being rendered as a paragraph. Ideally, the newlines should be ignored when displaying this text in the tooltip.

### Example

Given this code:
```fsharp
/// A type describing a unit in terms of its dimensions, base coefficient and offset, and recognized names
/// A unit of measure is uniquely defined by its dimensionality and its base coefficient and offset. Most units only
/// have a base coefficient, which defines the multiplicative conversion factor to scale the unit's value to the base
/// unit's equivalent. Absolute temperature units also have an additive base offset that scales the unit's value to the
/// base unit's zero-point. The `Names` property contains the list of strings that can be used to refer to the unit,
/// usually a combination of full names and abbrevations.
type Unit = {
    Dimensions: DimensionVector
    BaseCoefficient: float
    Names: string list
}
```
This is what appears in the tooltip. Note that this occurs even if the comment is wrapped in `<summary>`, `<remarks>`, or `<para>` tags.

image
(sorry for the poor quality screenshot; the Mac screenshot key combo interferes with the tooltip)

### Machine Info

* Operating system: **Darwin**
* Arch: **x64**
* VSCode: **1.25.1**
* Runtime: **.Net**
* Mono version:
```shell
Mono JIT compiler version 5.10.1.57 (2017-12/ea8a24b1bbf Tue Apr 24 14:53:01 EDT 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(3.6.0svn-mono-master/8b1520c8aae)
	GC:            sgen (concurrent by default)
```
* MSBuild version:
```shell
Microsoft (R) Build Engine version 15.6.0.0 (xplat-master/ca830585 Sun Mar 25 19:24:09 EDT 2018) for Mono
Copyright (C) Microsoft Corporation. All rights reserved.
```

MangelMaxime:
Fix #852

Markdown is space and whitespace sensitive. It will by itself remove undesirable blank lines, whitespaces, etc.

By triming the lines, the user can't format it's text correctly.

------------

Code:

```fs
/// A type describing a unit in terms of its dimensions, base coefficient and offset, and recognized names
/// A unit of measure is uniquely defined by its dimensionality and its base coefficient and offset. Most units only
/// have a base coefficient, which defines the multiplicative conversion factor to scale the unit's value to the base
/// unit's equivalent. Absolute temperature units also have an additive base offset that scales the unit's value to the
/// base unit's zero-point. The `Names` property contains the list of strings that can be used to refer to the unit,
/// usually a combination of full names and abbrevations.
/// * Maxime
///     * Maxime
///
/// ```fsharp
/// let maxime = ""
/// ```
type Unit = {
    Dimensions: obj
    BaseCoefficient: float
    Names: string list
}

```
Before:
img_1534

After:
img_4654

Now the output result is what user would expected when writing markdown. Indented list, etc.

### DIFF
--- a/src/Core/Utils.fs
+++ b/src/Core/Utils.fs
@@ -241,10 +241,6 @@ module Markdown =
     let createCommentBlock (comment: string) : MarkdownString =
         comment
         |> replaceXml
-        |> String.split [|'\n'|]
-        |> Array.filter (not << String.IsNullOrWhiteSpace)
-        |> Array.map String.trim
-        |> String.concat "\n\n"
         |> (fun v -> MarkdownString v)
 
 module Promise =


Krzysztof-Cieslak:
This breaks formatting for standard XML tooltips for, for example, framework functions
image

MangelMaxime:
Hum, ok then we need to find :

- a way to detect if this is xml based
- fix the `replaceXML` function

I will take a look at it

MangelMaxime:
Ok so seems like we can have 3 sources of doc comments:

- Markdown style
```fs
/// A type describing a unit in terms of its dimensions, base coefficient and offset, and recognized names
/// A unit of measure is uniquely defined by its dimensionality and its base coefficient and offset. Most units only
/// have a base coefficient, which defines the multiplicative conversion factor to scale the unit's value to the base
/// unit's equivalent. Absolute temperature units also have an additive base offset that scales the unit's value to the
/// base unit's zero-point. The `Names` property contains the list of strings that can be used to refer to the unit,
/// usually a combination of full names and abbrevations.
/// * Maxime
///     * Maxime
///
/// ```fsharp
/// let maxime = ""
/// ```
type Unit = {
    Dimensions: obj
    BaseCoefficient: float
    Names: string list
}
```
- XML style
```fs
/// <summary>Type of a formatting expression.</summary>
/// <typeparam name="Printer">Function type generated by printf.</typeparam>
/// <typeparam name="State">Type argument passed to %a formatters</typeparam>
/// <typeparam name="Residue">Value generated by the overall printf action (e.g. sprint generates a string)</typeparam>
/// <typeparam name="Result">Value generated after post processing (e.g. failwithf generates a string internally then raises an exception)</typeparam>
let p = ""
```
- Things comming from standard functions
```fs
let t x = List.filter
// When hovering `filter` I get:

// Returns a new collection containing only the elements of the collection
//  for which the given predicate returns "true"
// `list`: The input list.
// `predicate`: The function to test the input elements.
// Generic parameters:
// `'T` is `'b`
```

I guess option 3 is something coming out from [`TipFormatter`](https://github.com/fsharp/FsAutoComplete/blob/5d82a62347d3b0d96e48d32c1cd3bab9aa89df02/src/FsAutoComplete.Core/TipFormatter.fs) and I would say the output of the TipFormatter is either incorrect or should be improved.

@Krzysztof-Cieslak What do you think ?

MangelMaxime:
For now, I am able to support both XML and Markdown style comments.

Comparaison for the XML comments:

**Code:**

```fs
/// <summary>Extensible printf-style formatting for numbers and other datatypes</summary>
///
/// <remarks>Some random text goes here <c>Inline code</c>
///
/// The optional width is an integer indicating the minimal width of the
/// result. For instance, %6d prints an integer, prefixing it with spaces
/// to fill at least 6 characters. If width is '*', then an extra integer
/// argument is taken to specify the corresponding width.
///
/// Valid flags are:
/// <c>
///     0: add zeros instead of spaces to make up the required width
///     '-': left justify the result within the width specified
///     '+': add a '+' character if the number is positive (to match a '-' sign
///          for negatives)
///     ' ': add an extra space if the number is positive (to match a '-'
///              sign for negatives)
/// </c>
///
/// Here is an example of a link <a href="http://fable.io"></a>
/// <para>This should create a new paragraph</para>
/// The printf '#' flag is invalid and a compile-time error will be reported if it is used.</remarks>
/// <typeparam name="Printer">Function type generated by printf.</typeparam>
/// <typeparam name="State">Type argument passed to %a formatters</typeparam>
/// <typeparam name="Residue">Value generated by the overall printf action (e.g. sprint generates a string)</typeparam>
/// <typeparam name="Result">Value generated after post processing (e.g. failwithf generates a string internally then raises an exception)</typeparam>
/// <param name="Param1">Description of Param 1</param>
/// <param name="Param2">Description of Param 2</param>
/// <param name="Param3">Description of Param 3</param>
/// <returns>The concatenated string.</returns>
/// <exception cref="System.ArgumentNullException">Thrown when the input string is null.</exception>
/// <exception cref="System.ArgumentNullException">Thrown when the input string is null.</exception>
/// <exception cref="System.ArgumentNullException">Thrown when the input string is null.</exception>
let temp = ""
```

**Before:**


**Now:**


Still need to fix the **option 3**.

MangelMaxime:
@Krzysztof-Cieslak With a custom version of FSAC I get this result for the doc comment comming from XML docs of standard functions.



In theory, now all three sources of docs comment have the same formatting for the end user :) I will make a PR to the FSAC project tomorrow I think to discuss the changes.

PR #857, opened 2018-07-20
 Add notifications if project loading or parsing failed 

### LINKED ISSUE #827: Add notifications if project loading failed
This should be opt out with settings

MangelMaxime:
Fix #827


Krzysztof-Cieslak:
Couple of points:
1. We need info which project
2. There should be button to open project status view.
3. I don't think we need to notify for "not restored" - projects with such status should start automatic restore

MangelMaxime:
> 1. We need info which project

I do agree with that. I didn't find a nice way to show the project path. We can't put newline in an error message.

Would something like `Project loading failed: /path/to/my/fsproj/file` be ok ?

Also, is it ok to transform the `absolute` path into a `relative` one from the workspace `rootDir`?

> 2. There should be button to open project status view.

Ok will do.

> 3. I don't think we need to notify for "not restored" - projects with such status should start automatic restore

Ok will remove

Krzysztof-Cieslak:
I think it can be just `name.fsproj` without path at all - people usually don't have multiple projects with same name in one solution.

baronfel:
Agree with that. Plus if you have the button to take you right to the details screen, it mostly-eliminates the need for the full path. Because if you didn't have the button you'd have to go project file -> show load details yourself, at which point the full path would be useful.

Krzysztof-Cieslak:
Thanks a lot!

PR #862, opened 2018-07-25
Implement RecordStubGenerator and fix UnionCaseGenerator

### LINKED ISSUE #306: Record Stub Generation
The one thing I miss from Power Tools in VS

https://fsprojects.github.io/VisualFSharpPowerTools/recordstubgeneration.html

Is this possible?

MangelMaxime:
### ⚠️ Do not merge yet, this need a new version of FSAC ⚠️ 

Fix #306 

### Demo

This demo cover all the supported synthax:

## Union case generator

I fixed the Union case generator too because it was a problem in the `Pos` definition.
[Source](https://github.com/ionide/ionide-vscode-fsharp/compare/master...MangelMaxime:feature/record_generatio?expand=1#diff-626dfc57e4bcc2cd510c990760af4555R126)

### Demo

**Before**

**After**

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -1202,6 +1202,11 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Enables integration with FSharpLinter (additional warnings)"
+                },
+                "FSharp.recordStubGeneration": {
+					"type": "boolean",
+					"default": true,
+					"description": "Enables record stub generation"
 				},
 				"FSharp.unusedOpensAnalyzer": {
 					"type": "boolean",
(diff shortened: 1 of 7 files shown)


MangelMaxime:
One thing to note, the suggestion tooltip can only appear when `parse` has been done. This is because we need to know the AST in order to detect the case.

A visual indication is when you see the **red ripple**.

MangelMaxime:
Because this is listening to the cursor move event. I added a setting to disable the new feature.

Issue #858, opened 2018-07-20
Can't start in devMode


MangelMaxime:
I start FsAutoComplete from CLI:

```shell
➜  FsAutoComplete git:(master) ✗ dotnet src/FsAutoComplete.netcore/bin/Debug/netcoreapp2.0/fsautocomplete.dll --mode http
[21:32:52 INF] Smooth! Suave listener started in 40,563 with binding 127.0.0.1:8088
```

I set `devMode = true`.

And now, when `LanguageService` try to start it's failing.

```fs
doRetry startByDevMode
|> Promise.onSuccess (fun _ ->
    printfn "success"
    socketNotify <- startSocket "notify"
    socketNotifyWorkspace <- startSocket "notifyWorkspace"
    ()
)
|> Promise.onFail (fun _ ->
    printfn "Failed"
)
```

I only see `"Failed"` in the output. Any idea ?

Krzysztof-Cieslak:
Fixed by #868

PR #868, opened 2018-07-28
Fix ionide not starting when using FSAC in dev mode

Titaye:
doRetry logic for starting fsac process do not handle the case when childprocess is null in dev mode #858

### DIFF
--- a/src/Core/LanguageService.fs
+++ b/src/Core/LanguageService.fs
@@ -16,7 +16,7 @@ module node = Fable.Import.Node.Exports
 module LanguageService =
     let ax =  Globals.require.Invoke "axios" |> unbox<Axios.AxiosStatic>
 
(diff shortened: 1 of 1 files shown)


src/Core/LanguageService.fs
```diff
@@ -16,7 +16,7 @@ module node = Fable.Import.Node.Exports
 module LanguageService =
     let ax =  Globals.require.Invoke "axios" |> unbox<Axios.AxiosStatic>
 
-    let devMode = false
+    let devMode = true
```
Krzysztof-Cieslak: This shouldn't be committed ;-)

Krzysztof-Cieslak:
Thanks a lot!

Titaye:
I know... Stupid me...

Issue #887, opened 2018-08-18
Getting started UX


Krzysztof-Cieslak:
Replace current startup error handling with:

1. If user is on Windows and uses `net` and FSAC can't detect F# installation show message with button linking  to VS Build Tools
2. If user is on *nix and uses `net` and FSAC can't start show message with button linking to the mono
3. If user uses `netcore` and FSAC can't start show message with button linking to .Net Core installation

Possibly add:
1. If user is on `net` but has `dotnet` installed suggest using `netcore`. Button for opening the settings view, ignoring suggestion once, and ignoring suggestion always
2. If user is on `netcore`, and `dotnet` is not installed suggest using `net` (if `mono` is installed). Button for opening the settings view, ignoring suggestion once, and ignoring suggestion always

--- 

Update ionide.io: 
1. include latest ways of installing F#. 
2. Mark `dotnet` as required dependency (not true at the moment, but I want to convince people to use it, to make potential future change easier)

PR #888, opened 2018-08-19
WIP: Getting started ux

baronfel:
This is built on top of #886 but begins to implement the changes described in #887.

Example notification when fsc/msbuild are not found for windows:
image

Example notification when fsc/msbuild are not found for unix:
image

Remaining Tasks:
- [x] `netcore` mode when `dotnet` cannot be found (will require making finding of `dotnet` optional)
- [x] suggest `netcore` if user is on `net` and mono binaries are found
- [x] suggest `net` if user is on `netcore` and `dotnet` cannot be found, but mono binaries are found


baronfel:
I've filled out more of the TODOs, but testing locally is proving difficult because for some reason despite my local workspace settings set to `FSharp.fsacRuntime: net`, the `targetRuntime` logic is returning `netcore`

Issue #97, opened 2016-07-20
Rename doesn't work on double-ticked identifiers


Krzysztof-Cieslak:
e.g. 

``` fsharp
``Num Of Files    ``
```

![](http://i.imgur.com/0LDsaUY.png)

Krzysztof-Cieslak:
I **think** it can be solved by using `languages.setLanguageConfiguration` and creating good RegEx for wordPattern in [LanguageConfiguration](https://code.visualstudio.com/docs/extensionAPI/vscode-api#LanguageConfiguration)

vasily-kirichenko:
@Krzysztof-Cieslak where that `wordPattern` should be set? I've searched the codebase and have not found any place where LanguageConfiguration is set.

Krzysztof-Cieslak:
In latest Code release they've added option to set wordPattern in `language-configuration.json` - https://github.com/ionide/ionide-vscode-fsharp/blob/master/release/language-configuration.json

MangelMaxime:
I tried to solve this issue but even if I got a regex matching the correct groups VSCode doesn't seems to work as expected.

If someone want to give it a try or confirm my bug [here is a gist](https://gist.github.com/MangelMaxime/7b76a694c9704fbc29de4bbb14e01197) with the regex and test strings.

MangelMaxime:
For reference, I opened an issue on VSCode repo.

https://github.com/Microsoft/vscode/issues/32762

Krzysztof-Cieslak:
Closing it here as an upstream VSCode issue. I hope it will be solved with new API in renameProvider (https://github.com/Microsoft/vscode/pull/37691)

PR #891, opened 2018-08-20
Improved the word pattern regex. Thanks to @MangelMaxime (#97)

mvsmal:
I updated the word pattern regexp to support double ticked symbols renaming as @MangelMaxime proposed.
Unfortunately it works only on the first word if there are more than one. It seems that VSCode applies the regexp starting from the beginning of the word only.


mvsmal:
The current behavior:
double-tick-rename

And with this fix:
double-tick-rename-new

PR #892, opened 2018-08-20
clean up types

selketjah:
refactor types in DTO file to F# guidelines


selketjah:
@Krzysztof-Cieslak Any specific reason some types are lower case?

If no, can I fix that?

selketjah:
ready to merge
will fix the lower case if need be in next PR

selketjah:
@Krzysztof-Cieslak  This can be merged back already.

Regarding the question of lower case types, I will handle that in a different PR should there be no reason they are lower case :)

PR #895, opened 2018-08-22
Normalize leading space for docs tooltips

### LINKED ISSUE #885: XML comments not formatted anymore
## Description

I am currently on a Windows machine, but I have the same problem with latest VS Code + latest Iionide on macOS.

Previously this code would have nicely formatted my comments:

```fsharp
module Say =
    /// ** Description **
    /// This is a test description
    /// ** Parameters **
    ///     - name: The name to say hello to.
    /// ** Output **
    /// Prints "Hello {name}" into the console.
    let hello name =
        printfn "Hello %s" name
```

but now it doesn't line break between `**Description**` and the new line and it also doesn't format it in bold anymore when I `**` something.

This certainly worked before, because initially I didn't have such XML comments and only after seeing them in a different F# repository and seeing how they created nice tooltips I started to adopt the same style to all my code.

## Current result:

image

## Steps to reproduce:

1. Download latest VS Code
2. Download latest Ionide
3. `dotnet new classlib -lang F#`
4. Copy the code from above

MangelMaxime:
Fix #885

Before:

Now:

### DIFF
--- a/src/Core/Utils.fs
+++ b/src/Core/Utils.fs
@@ -268,9 +268,22 @@ module Markdown =
             res.Replace(oldValue, newValue)
         ) res
 
+    let private normalizeLeadingSpace (content : string) =
+        content
+            .Replace("\r\n", "\n")
+            .Split('\n')
+        |> Array.map(fun line ->
+            if line.Length > 1 && line.[0] = ' ' then
+                line.[1..]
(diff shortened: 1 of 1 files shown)


dustinmoris:
Awesome, thank you!

PR #916, opened 2018-09-04
Fix #913: Add .fable as excluded dir by default

### LINKED ISSUE #913: Should we prevent Ionide to load files under .fable folder
I am opening this here because I don't think there is a benefit to parsing/loading these projects in Ionide/FSAC but I can be wrong.

cc @alfonsogarciacaro 

If we add this feature, I propose to make it configurable via a setting.

IMO it's like for projects under `paket-files`, `packages` I don't think there is a benefit to loading them but I can be wrong of course :)

alfonsogarciacaro:


Alex2357:
Seems the fix doesn't always work. I have this issue in 3.33 https://github.com/elmish/react/issues/36

PR #943, opened 2018-10-19
fix file watchers doesnt consider FSharp.excludeProjectDirectories

enricosada:
the setting `FSharp.excludeProjectDirectories` configure the ignored dirs
This was not considered by file watchers for changed/creation/deletion, so updating in these dir will trigger a reload/init workspace

- do not load created/changed projects/sln from ignored dirs
- do not reload project.assets.json of ignored fsproj


Krzysztof-Cieslak:
Thanks!

PR #963, opened 2018-11-16
Remove CodeOutline feature

vbfox:
VSCode has a native one and it work well with F# code, I'm not sure that Ionide need a separate one.

(Simply a proposal because I have no idea if it's a good idea but 🤷‍♂️ )


Krzysztof-Cieslak:
Yeah, it is one of the things I was planing to do. LGTM

PR #964, opened 2018-11-16
Enable LineLens by default when CodeLens is disabled

vbfox:
Hopefully it will make more people notice the feature.

I also changed the default colors, I like ones that aren't very contrasted with the background but depending on screen calibration it's hard to read and some people made the remark to me.

The new colors have been choosen to have a ~3.0 contrast ratio with the default VSCode theme (WCAG AA minimum for large text)

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -1280,7 +1280,7 @@
 						"replaceCodeLens",
 						"always"
 					],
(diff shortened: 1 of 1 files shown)


Krzysztof-Cieslak:
Thanks!

PR #979, opened 2018-12-14
[WIP] Try to restore restore if project loading fails

forki:
image

### DIFF
--- a/src/Core/LanguageService.fs
+++ b/src/Core/LanguageService.fs
@@ -258,20 +258,26 @@ module LanguageService =
             Data = { res.Data with
                         Info = parseInfo(res.Data.Info) } }
 
-    let project s =
+    let rec tryProjectLoading restore s =
         { ProjectRequest.FileName = s }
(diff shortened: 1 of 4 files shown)


forki:
Oh I was hoping for a CI build to kick in. build.cmd fails for me locally

src/Core/LanguageService.fs
```diff
@@ -258,21 +258,27 @@ module LanguageService =
...
+                vscode.window.showErrorMessage(msg, "Show status")
+                |> Promise.map(fun res ->
+                    if res = "Show status" then
+                        Preview.showStatus s (path.basename(s))
+                        |> ignore
+                )
+                |> ignore
         )
 
+    let project s = tryProjectLoading restore s
```
Krzysztof-Cieslak: `restore` is undefined?

Krzysztof-Cieslak:
I think this needs a bit different approach and modifying current automatic restoring rather than new mechanism in this place. Closing this PR, but I'm looking at the issue right now.

PR #994, opened 2019-01-25
Fix #921: Add smart indent support

### LINKED ISSUE #921: Automatic indent support.
## Problem with Indenting
I come from Python community and just started learning F#. One thing that struck me was not able to auto-indent after specific expression/assignment. Like after module, let expression or simply after `=` . 

### Proposed Solution
Some level of auto-indenting would be great. I know it is hard to tell where an expression ends, but we can surely tell where it starts and indent any newline cursor to that. 

## Problem with Un-Indenting
I had to use backspace four time before moving to another block of code. It is definitely a part where editors can support and make indented languages better.

### Proposed Solution
It would be better I can un-indent with a single stroke of backspace. Since we do not write whitespace at the start of the line, it will be easier to differentiate within/out of expression. 

## Problem with lot of spaces
I know that there is auto-formatting available. But it will really be nice to cleanup any extra whitespaces and newlines in the file. 

Please note, I am just a noob here and expect to get the same feeling typing Python code in a Python editor.

MangelMaxime:
Update the language configuration on the fly and notify the user:

Code wrote with smart indent enabled:

Code wrote with smart indent disabled:

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -1317,7 +1317,12 @@
 					],
 					"default": "sameAsFileExplorer",
 					"scope": "window"
-				}
+                },
+                "FSharp.smartIndent": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Enables smart indent feature"
+                }
 			}
 		},
 		"jsonValidation": [
--- a/src/Components/LanguageConfiguration.fs
+++ b/src/Components/LanguageConfiguration.fs
@@ -7,7 +7,17 @@ open Ionide.VSCode.Helpers
 
 module LanguageConfiguration =
 
-    let activate (context : ExtensionContext) =
+    let mutable private reference : Disposable option = None
+
+    let indentationRules =
+        jsOptions<IndentationRule> (fun o ->
+            o.increaseIndentPattern <- JS.RegExp.Create("""^(\s*(module|type|let|static member|member)\b.*=\s*)$|^(\s*(with get|and set)\b.*=.*)$|^(\s*(if|elif|then|else|static member|member)).*$""")
+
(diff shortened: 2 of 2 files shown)


jschiefer:
typeClass 🤣

inosik:
Doesn't seem to work for me, even after restarting VS Code.

MangelMaxime:
Did you set `FSharp: Smart indent` and `Editor: Auto Indent`to true?

inosik:
My bad, I disabled "Auto Indent" long time ago, because I thought it would be annoying when moving code around 😅

It's nice when you're typing new code, but moving existing code around isn't working properly:



I don't think we can fix this in Ionide.

MangelMaxime:
Smart indent is kind of really hard to set up in Ionide because we are working on a language that is indentation based.

In other languages, it's really easy to determine the block. Because in general, you can identify them with opening and closing symbols: `{}` or `;` or `[]`, etc.

When working on the F# smart indent, I took the decision to make it minimal and auto-indent where it as the best chances to be "correct".

For example, when creating a `module MyModule =` the code is always indented.
When defining a `let x =` the code after that is generally indented.

When I compare my experience with or without smart indent it's pretty similar when creating code. Before smart-indent, I had to indent myself to the right.
Now, it's always indented to the right but I sometimes need to de-indent.

In your gif, I am not sure why `fetchGif` is being indented to the right. This is VSCode determining that it should but I would expect it to not change.

Perhaps this is a bug/limitation on their side?

Also, I didn't use yet smart-indent on a daily I only tested it when creating the feature but thanks for the reminder I will activate it :) And will see if we can improve things or if there are bugs etc. :)

Issue #1002, opened 2019-02-01
Unable To Open Project - D:/Programowanie


paytonrules:
On opening a folder with two fsharp projects in it ionide detects both of them, but fails to load with the error below: 

<img width="240" alt="nerdstrengthtests_assemblyinfo_fs_ _nerdstrengthunity" src="https://user-images.githubusercontent.com/21752/52123789-c3dc1200-25ec-11e9-9442-3ebe7d334bfa.png">

It looks like a developer path is hard coded somewhere? I'm on MacOS so I don't even have a d drive, let alone that path.

paytonrules:
I should add - the project builds just fine at the command line with `dotnet build`.

Krzysztof-Cieslak:
Huh, it doesn’t look like - https://github.com/ionide/ionide-vscode-fsharp/search?utf8=✓&q=programowanie&type=

Krzysztof-Cieslak:
Can you open F# solution explorer (F# icon in the activity bar), right click the failing project and then check project status?

paytonrules:
I can't - not for lack of trying. I just get that same error again: "Unable to open 'Project NerdStrengthTests status': D:/Programowanie/ionide/src/Components/SolutionExplorer.fs." 

I'm kinda shocked the unique directory path doesn't immediately make a contributor say "oh no that's my machine!"

enriqueMartinezInter:
Same thing happening to me
![image](https://user-images.githubusercontent.com/12499680/52314450-50cdf500-2978-11e9-825b-2660fd7998f1.png)

MangelMaxime:
@Krzysztof-Cieslak `Programowanie` seems to a polish word meaning `programming` (but you know better than me ^^).

Perhaps, one of the dependencies of Ionide has been compiled on a Polish computer and have `Programowanie` in hard in the code or the path resolution functions.

Krzysztof-Cieslak:
> Perhaps, one of the dependencies of Ionide has been compiled on a Polish computer and have Programowanie in hard in the code or the path resolution functions.

Yeah, sure. But it has nothing to do with the bug happening here - every unhandled exception will have some path displayed. And yeah probably our webpack config has some small issue (it should replace those paths with relative)... but it’s not connected to behavior in any way

MangelMaxime:
Ah yes, I see what you mean.

And you are right #1008 will fix this issue. I didn't read correctly the thread and didn't understand that the error was coming when trying to show the project status...

[...]

PR #1008, opened 2019-02-06
Upgrade to VS Code's webview API and fix show project status

### LINKED ISSUE #958: Upgrade to VS Code's webview API
Hi, I'm on the VS Code team. I noticed that your extension uses the `vscode.previewHtml` command which is deprecated and which we are actively working to remove: https://github.com/Microsoft/vscode/issues/62630

We've developed a [webview API][docs] that provides a much better developer experience and offers a number of important security and compatibility benefits over `previewHtml`. We cannot fix `previewHtml` without breaking backwards compatibility, and have instead opted to remove it. 

To ensure that your extension continues to work properly in VS Code, please try upgrading to use the new Webview API. You can find documentation on the API usage [here][docs]. Let me know if you have any questions or concerns about this migration

[docs]: (https://code.visualstudio.com/docs/extensions/webview

### LINKED ISSUE #1003: Show project status seems broken
I am fixing it when working on https://github.com/ionide/ionide-vscode-fsharp/issues/958

MangelMaxime:
This fix #958 and fix #1003

Implementation has been adapted from [webview-sample](https://github.com/Microsoft/vscode-extension-samples/blob/master/webview-sample/src/extension.ts)

This PR needs https://github.com/ionide/ionide-vscode-helpers/pull/27 to merged before. I will update the dependencies when ionide-vscode-helpers will be updated.

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -17,7 +17,7 @@ git https://github.com/fsharp-editing/Forge.git master build:"build.sh Build", O
 
 git https://github.com/ionide/ionide-fsgrammar.git
 
-github ionide/ionide-vscode-helpers src/Fable.Import.VSCode.fs
+github MangelMaxime/ionide-vscode-helpers:webview_api src/Fable.Import.VSCode.fs
 github ionide/ionide-vscode-helpers src/Helpers.fs
 github ionide/ionide-vscode-helpers src/Fable.Import.Axios.fs
 github ionide/ionide-vscode-helpers src/Fable.Import.ws.fs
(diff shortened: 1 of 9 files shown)


Krzysztof-Cieslak:
It probably also fix #1002

Krzysztof-Cieslak:
I’ve merged the helpers PR. Awesome work, thanks a lot!

MangelMaxime:
About #1002 I don't know. 

The fix included in this PR, is only for the `Show project status` feature. Because the following [lines](https://github.com/ionide/ionide-vscode-fsharp/pull/1008/files#diff-03adafec899d1868b2e173596f3ae7c9R648) is called only when generating the HTML code.

I think the function which generates the `files` list is adding `Folder` and `File` into it. And because the `files` is of type `Model list` it's a valid representation. I think we have an issue in the code that generates the "Workspace" tree.

PR #1011, opened 2019-02-07
Feature/fix get help

### LINKED ISSUE #920: Unable to open 'dev15.query': resource is not available.
On archlinux, vscode 1.27.2 and Ionide-fsharp 3.27.0, when I run <code>F#: Get Help</code> I got the following error:
 Unable to open 'dev15.query': resource is not available.

MangelMaxime:
Fix #920

The new docs seems to use `localStorage` in order to decide and store which language to display.

So by default, the user will always see the C# version of the documentation. Until he manually chooses to set F# as the display lang.


Krzysztof-Cieslak:
Hey @cartermp any idea if there is a way to choose programming language with URL params for new docs? And if not... who should we convince to make it happen?

cartermp:
Nope, no current way. I recommend filing a bug here: https://github.com/MicrosoftDocs/feedback

Additionally, the current behavior (which I've called out as not correct) just sort of blindly goes with whatever the next item in the list is if there aren't samples for a given language. So you'll see examples in C++ if F# is the default, but there aren't F# samples. I asked that this default to C# instead, but so far that doesn't appear to be the case.

Issue #1026, opened 2019-02-18
Type Provider issue when using netcore FSAC


isaacabraham:
Using the netcore version of FSAC doesn't appear to work with type providers. Using it with e.g. JSON TP gives this error:

![image](https://user-images.githubusercontent.com/1781813/52953415-b2825c00-3387-11e9-9eb0-693acb14614a.png)

I've tried adding a reference to this System.Private.CoreLib as a #r but no change. Any ideas?

Krzysztof-Cieslak:
Cc: @enricosada.

Also CC: @MangelMaxime - it seems that formating in tooltip in above screenshot is wrong. We probably should create separate issue for that

MangelMaxime:
If an issue is opened with a reproduction code I can take a look at fixing the tooltip.

enricosada:
tracking this issue in https://github.com/fsharp/FsAutoComplete/issues/348

isaacabraham:
@Krzysztof-Cieslak for the moment I'd like to suggest removing the pop-ups to move over to the netcore FSAC - it's actually not necessarily good advice, since it doesn't work with type providers. Or at the very least it should be able to stop the alert appearing - it currently shows *every single time* that you start Code.

baronfel:
I think that makes a lot of sense. I think we could easily make it so if there was a user-set value for the fsacRuntime then we'd just ignore the suggestion overall.

isaacabraham:
@baronfel have just updated to latest ionide but the messages are still showing every time we start.

isaacabraham:
@baronfel @Krzysztof-Cieslak do you know of any reason why this might not be working? Do I have to first explicitly move to netcore and then back to net before it takes effect or something?

drk-mtr:
I had issues with this too recently when using the SQLProvider type provider. I read a comment that suggested the wrong version of FSharp.Data or the compiler might be getting used, and that wiping the `.vscode` folder from the your windows user folder, and the `Code` directory from `AppData` might be a solution - this seemed to work for me but I also cleaned my paket cache at the same time so I can't be certain.

Great to see the prompts to switch to netcore each time are being removed! Prompts generally infuriate me, one of the main reason I switched to vscode is that some other apps can be quite "noisy" in terms of the development experience, so the less popups the better in my view.

drk-mtr:
[...]

PR #1045, opened 2019-03-09
tweak the configuration matrix to stop bugging users about their life choices

baronfel:
Addresses a concern raised in #1026 about popups for target-runtime suggestions being raised even when the user has set an explicit value. Now we don't do that anymore.

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -1075,7 +1075,6 @@
 			"properties": {
 				"FSharp.fsacRuntime": {
 					"type": "string",
-					"default": "net",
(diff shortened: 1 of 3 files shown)


baronfel:
I also added the ability for the 'suggest' messages to let you click a button to set the different runtime, and prompt the user to reload after runtime changes are set.

baronfel:
without a runtime set:
image
image


with a runtime set there's no prompt

Krzysztof-Cieslak:
LGTM

PR #1047, opened 2019-03-10
restart all of the project-load loop if fasc crashes and is reborn

### LINKED ISSUE #1046: Projects aren't reloaded when FSAC is restarted
This is kind-of a follow-on from #832, in that #832 restarted the FSAC process, but we don't re-input the workspace or project load commands necessary to ensure the currently-loaded files are all parsed.  So the requirement here would be that

* when FSAC is restarted
* the current workspace/project selections are reloaded into the new FSAC instance.

I thought that this could be easily implemented through use of an event in the LanguageService module, and this event would be triggered once each time the FSAC instance was started.
Then, in fsharp.fs, the portion of the initial ionide startup that is related to workspace peek, workspace selection, and project loads would be extracted out into a separate function that could be invoked on the event handler.  I tried this locally but am not sure that this is a good way/the best way to get to the desired result.

Some components in the startup don't directly need to be invoked each time FSAC runs, so they could be left on the current initialization pathway. This would be things like tooltips, quick fixes, etc.  They just send requests to FSAC, so as long as we are able to repopulate FSAC with parsed projects on a reload, they should continue to 'just work'.

baronfel:
This fixes #1046 by adding an event that's fired when FSAC is restarted, and hooks up the workspacepeek/project load cycle to that event instead of being invoked only once.

Many tools were moved out of the project-finding loop as part of this as well, because they don't need to be re-registered multiple times.

I'm not certain this is the best way, but it seems to be working for me right now. I'd appreciate if someone else could take a test drive too.

### DIFF
--- a/src/Core/LanguageService.fs
+++ b/src/Core/LanguageService.fs
@@ -719,10 +719,15 @@ module LanguageService =
             | _ -> (VSCode.getPluginPath "Ionide.Ionide-fsharp")
         spawnFSACForRuntime targetRuntime ionidePluginPath
 
+    let private fsacStarted = new FSharp.Control.Event<unit>()
+
+    let Ready = fsacStarted.Publish
+
     let start () =
         let rec doRetry procPromise =
             procPromise ()
             |> Promise.onSuccess (fun (childProcess: ChildProcess.ChildProcess) ->
+                log.Info("FSAC started")
                 childProcess.on("exit", fun () ->
                     if exitRequested
                     then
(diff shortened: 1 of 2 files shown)


Krzysztof-Cieslak:
I'm bit afraid that moving registration around will break some hidden assumptions about the order of activation of things...

Krzysztof-Cieslak:
Let's not do that right now - things may change too much with incoming LSP version of FSAC

PR #1049, opened 2019-03-11
allow users to set a configurable FSAC url

### LINKED ISSUE #1048: Users should be able to configure the FSAC url
The use case here would be a scenario where a user has

* multiple FSAC instances and wants to choose one to test against (potentially for FSAC version behavior comparison?)
* work hosted on another machine/vm (for example source code in a docker container?) and the user wants to work 'remotely'

So then the user could set a config like `FSharp.FSACUrl` to `https://my-vm-hostname:<FSAC_PORT>` and Ionide would use that url instead of spawning a process on the local machine.  This would also allow easier debugging while working locally without having to set the `debug` flag to `true` and recompiling.

baronfel:
Fixes #1048 by adding a new configuration setting and flowing that through.  If the url is set we don't spawn an FSAC instance, but if it's not we use the same logic as before.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,13 +140,15 @@ The new extension window will appear with window title `Extension development ho
 
 ### Working with FSAC
 
-1. Run `build.cmd Build` \ `build.sh Build`
-2. Open Ionide-vscode-fsharp in VSCode.
-3. Run FSAC:
-    - or from sources in your favorite editor: open the `FsAutoComplete.sln`, start debugging with  `.NET Core mode http (debug)` configuration
+1. Run FSAC:
+    - or from sources in your favorite editor: open the `FsAutoComplete.sln`, start debugging with  `.NET Core mode http (debug)` configuration (default to port `8088`)
     - or run it with `--mode http --port 8088` arguments
-4. Press F5 in VSCode to build Ionide and start experimental instance
-5. In the new instance, set the `FSharp.fsacUrl` setting to `http://127.0.0.1:8088` and reload the experimental instance. Note that if you have changed the standard FSAC configuration to use a different port you will need to update that port number in the `FSharp.fsacUrl` setting.
+2. In vscode, set the `FSharp.fsacUrl` setting to `http://localhost:8088`
+3. Reload the instance with `> Developer: Reload Window`.
+
+The port number can be changed as needed.
+
+Remove that setting to go back to FSAC bundled in Ionide extension.
 
 ### Dependencies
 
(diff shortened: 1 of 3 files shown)


Krzysztof-Cieslak:
Can you rebase?

baronfel:
A note here: The default for the fsacRuntime config value needed to be removed because VSCode would silently add it on `Configuration.get` calls, so we couldn't tell the difference between user-set and the default value. So now we handle the default-setting in code in the language service.

Krzysztof-Cieslak:
One thing I’d add to it would be some kind of notifications when we can’t connect to FSAC pointing to the setting - people will mess their setting and they will forget they’ve changed something and then they will complain that F# is broken ;-)

So in our request helpers I’d check for the particular error (404? Timeout?) and display error message

baronfel:
That sounds reasonable to me. I won't be able to get to it until after Thursday (prepping for Meetup ATM) but I'll add that and we can take another look then.

enricosada:
That's a really interesting feature, will be awesome to have for the current Outreachy internship on FSAC (it's already hard to explain FSAC alone).
I too missed it, having to run vscode in debug just to check FSAC locally.

So i fixed the missing thing, i hope it helps @baronfel (good luck for meetup!)

@Krzysztof-Cieslak afaik is now ready, i think i added the feedback.

- rebased on latest master
- closed/reopen to trigger azure devops build
- updated the instruction on CONTRIBUTING
- fixed the logic to start FSAC (the name of config was different `fsacUrlConfigured = Configuration.tryGet "FSharp.FSACUrl"` vs `Configuration.tryGet "FSharp.fsacUrl"` so wasnt working correctly)
- added the notify at startup for users, if that config is used

image

baronfel:
Thank you @enricosada! I hope it helps :) I've got a coworker at work who like to dev on windows using ionide but has to mount the source on osx for reasons, and he was the key driver for this feature.

CONTRIBUTING.md
```diff
@@ -137,14 +137,16 @@ Once the initial build on the command line is completed, you should use vscode i
...
-1. Open Ionide-vscode-fsharp in VSCode.
-2. Set `devMode` to `true` in `src/Core/LanguageService.fs`
-3. Open FSAC in VS
-4. Start FSAC.Suave in VS
-5. Press F5 in VSCode to build Ionide and start experimental instance
+2. Open Ionide-vscode-fsharp in VSCode.
+3. Run FSAC:
+    - or from sources in your favorite editor: open the `FsAutoComplete.sln`, start debugging with  `.NET Core mode http (debug)` configuration
+    - or run it with `--mode http --port 8088` arguments
+4. Press F5 in VSCode to build Ionide and start experimental instance
```
Krzysztof-Cieslak: You don't really need an experimental instance in this case. Just set the setting in any instance. This feature is enabling debugging FSAC without building Ionide at all, it may be worth to emphasise this in this guide.

enricosada: done

enricosada:
updated doc, no need for experimental instance

Krzysztof-Cieslak:
Thanks a lot to both of you.

PR #1051, opened 2019-03-14
use Fake4

enricosada:
- no deprecation warnings
- target net, less dependencies needed

require rebase after #1050 is merged (ci green https://dev.azure.com/enricosada/ionide-vscode-fsharp/_build/results?buildId=65)


MangelMaxime:
Why not upgrade to FAKE 5 API?

Except for the "less dependencies needed" part which of course would be the same as now. By upgrading, to FAKE5 we would use the stable and maintained version of FAKE. I don't know if FAKE4 is still getting support.

enricosada:
Because is easier to do atm, i just want to avoid the deprecation warnings.

Do do that i just need to restrict the FAKE package version. atm works, i want to push others changes, no time to upgrade

Krzysztof-Cieslak:
TBF, I don't like this - the only advantage is removing of warnings... which doesn't sound as good justification for the change.

enricosada:
Removing lots of warning is the reason. Is a wall of text in the build output. 

I’ll close so, no personal plan to upgrade, no time for that

PR #1054, opened 2019-03-15
add `experimental-fsharp` extension

enricosada:
Add build of a new `experimental-fsharp` extension, who can be used to test new features or changes.

- the `experimental-fsharp` extension is built in `release-exp` directory
- add a compiler define `IONIDE_EXPERIMENTAL` when built in experimental extension
- add vendor lib with another copy of fsac, currectly built from tag `vnext` instead of `master`
- use assets (package.json, images, etc) of `ionide-fsharp`, but replace identifiers (done in code under `IONIDE_EXPERIMENTAL` define)
- add `RELEASE_NOTES_EXPERIMENTAL.md` and `README_EXPERIMENTAL.md`
- set experimental extension as preview

## Build infra
- add `BuildPackageExp` to build just the experimental package, and related `ReleaseExp` target to release it
- internal fake targets have `Exp` prefix
- azure pipeline build both extensions in parallel on different jobs
- both stable and experiemental have prerelease numbers in PR/branch from azure devops ci, like `PR{PR_NUMBER}-{UNIQUE_BUILD_NUMBER}` => `Ionide-fsharp-3.35.0-PR1054.47.vsix`. Users can download from build job summary and install as vsix

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,4 +75,4 @@ release/package-lock\.json
 /release/bin_netcore/
 /.ionide
 /vendor/.paket/Paket.Restore.targets
-/release-vnext/
+/release-exp/
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,5 @@ release/package-lock\.json
 *.vsix
 /release/bin_netcore/
 /.ionide
+/vendor/.paket/Paket.Restore.targets
+/release-vnext/
--- a/README_EXPERIMENTAL.md
+++ b/README_EXPERIMENTAL.md
@@ -1,10 +1,10 @@
-# [Ionide-VSCode: FSharp - Experimental](https://marketplace.visualstudio.com/items/Ionide.Ionide-fsharp-experimental)
+# [Ionide: Experimental FSharp](https://marketplace.visualstudio.com/items/Ionide.experimental-fsharp)
 **Enhanced F# Language Features for Visual Studio Code (EXPERIMENTAL VERSION)**
 
 _Part of the [Ionide](http://ionide.io) plugin suite._ Read detailed documentation at [Ionide docs page](http://ionide.io/docs).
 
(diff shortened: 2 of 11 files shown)


baronfel:
As another alternative, prerelease versions of ionide could be published, and the extensions menu in VSCode allows for explicit version selection, which o believe includes prerelease versions. What do you think of that?

enricosada:
To summarize, @Krzysztof-Cieslak can add more.

This is not for do a nightly.
Atm will be used to test ionide with FSAC vNext (a big change for project/sln/fsx loading), i dont want to bundle 4 FSAC binaries under a flag, or create issue for users.

Why another extension id:

- vscode marketplace doesnt support prerelease or other channels. no beta/preview.
- Personally i do not want to make ppl install by vsix. too much effort, risk of stale version for bug report, risk of broken upgrades/uninstall, i want auto update. Personally i'll not use it if is a vsix, too much work to do some testing. i'd like to lower the barrier.
- idea is to use this with Visual Studio Code Insider only (the green vscode), so you have both normal vscode blue (stable) and green (vnext) and easy to test both on same directories (zero effort, just open one and after the other).

baronfel:
Makes sense to me! I look forward to trying it out.

Krzysztof-Cieslak:
Maybe we should change the name to `Experimental` from `vNext`? `vNext` kinda suggest this is nightly/recent good build/something like that while in our case it's not necessarily that - it's rather a playground for things that are potentially breaking.

enricosada:
Good idea. I’ll change that

enricosada:
Atm build ok, install ok, but after two vscode reload, the extension installed from vsix  get replaced with normal `Ionide-fsharp` extension updated from gallery.

image

I dont have clue why does that, the vsix doesnt contains any reference to `ionide-fsharp`, and i changed also the name to `experimental-fsharp` to be sure.

enricosada:
Ok got it, the vsix has an id and a guid too.
Replaced both.

Also set experimental as preview

enricosada:
@Krzysztof-Cieslak ok ready to review.

the `experimental-fsharp` extension can be downloaded from [job summary artifacts](https://dev.azure.com/ionide/Ionide-VSCode-FSharp/_build/results?buildId=71&view=results)

enricosada:
Ok now both stable and experiemental have prerelease in PR (so like `Ionide-fsharp-3.35.0-PR1054.47.vsix`).
So user can just download from build job summary and install as vsix

PR #1055, opened 2019-03-20
refactor extract function to get ionide plugin path

enricosada:


enricosada:
trying to minimize diff of https://github.com/ionide/ionide-vscode-fsharp/pull/1054

PR #1056, opened 2019-03-20
remove workaround, already fixed in FSAC

enricosada:
fixed in https://github.com/fsharp/FsAutoComplete/commit/837ec3145b5bb7d9b0c4f7327bbee0c7fa0965ba

ref https://github.com/fsharp/FsAutoComplete/blob/5a217a4a1e08a8f1c27e30a0b5c4cdab7abab586/src/FsAutoComplete/FsAutoComplete.fsproj#L45-L54

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -103,7 +103,6 @@ Target "CopyFSAC" (fun _ ->
 
 
     CopyDir releaseBin fsacBin (fun _ -> true)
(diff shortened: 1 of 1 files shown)


enricosada:
@Krzysztof-Cieslak 	i'll incorporate the workarounds added to ionide in FSAC directly.
FSAC should work OOTB, with just `bin/release` and `bin/release_netcore` binaries.

This workaround was already fixed time ago.
As a note, the FSAC test suite was failing correctly, but tests were disabled. Now all tests are enabled.

PR #1059, opened 2019-04-02
remove workarounds

enricosada:
remove add default manifest workaround.
fixed by https://github.com/fsharp/FsAutoComplete/commit/9a35503f78382059b0efbfe3a53ea44c73a9629b

remove copy `.deps.json` and `.runtime.config.json` workaround
the `.dev.json` is not added, because is used only for development not in production (contains local paths from build machine)
fixed by https://github.com/fsharp/FsAutoComplete/commit/bd5ad65b939dac93c4bb39aa2ccb4d526b32d7e2


enricosada:
@Krzysztof-Cieslak i double checked the generated vsix, contains the right files

Issue #1066, opened 2019-04-12
"File <xyz>.fs can't be found in any parsed project. Usually .fs files should be included in .fsproj file" cannot be disabled


ollipekka:
**Describe the bug**
The notification that file is not in the project file cannot be disabled. This is obstructs users who keep file lists separately to support legacy & netstandard projects and include them to the project through `<Import Project=".."/>`. 

It probably doesn't make sense to support this use case for project parsing, but it would be nice to be able to disable the notification.

**To Reproduce**
1. Create file MyFile.fs to existing solution.
2. Open MyFile.fs & observe the notification to show up.
3. Try to disable the notification from the UI
4. Open some other file
5. Open MyFile.fs & observe the notification to show up again.

**Expected behaviour**
After disabling the notification, the notification should not reappear.
**Screenshots**

**Environment (please complete the following information):**
 - OSX
 - Ionide version: 3.35
 - VSCode version: 1.33.1
 - dotnet SDK version: 2.1.505
 - mono / .Net Framework version: 5.16.1.0  (mono)

**Additional context**

enricosada:
sent PR #1070

PR #1070, opened 2019-04-14
add missing config for FSharp.notifyFsNotInFsproj

enricosada:
without the config setting the UI disable doesnt work because cannot store the preference

fix https://github.com/ionide/ionide-vscode-fsharp/issues/1066


Krzysztof-Cieslak:
🤦‍♂️

PR #1072, opened 2019-04-14
add fsdn command

enricosada:
use new `fsdn` command of FSAC.

add `F#: search by signatures on FSDN` command to allow query https://fsdn.azurewebsites.net website

open a picker to choose the result, and add it to the current caret


enricosada:
@Krzysztof-Cieslak ready to review

PR #1074, opened 2019-04-15
allow to build stable version of experimental

enricosada:
disable prerelease tag if `ionide.isExperimentalRelease` var is set in azure devops build


enricosada:
ready

PR #1085, opened 2019-04-30
Add Info Panel feature

Krzysztof-Cieslak:
InfoPanel

I'd be happy to hear feedback/thoughts on the UX of this feature - should it be updated on cursor move? on mouse hover (to replace tooltips)? Both? Anything else?


Krzysztof-Cieslak:
(Also, I really wish that Code to have an actual concept of "panel", it would make it much nicer)

jcmrva:
Love it. Cursor move / mouseover config option.

Krzysztof-Cieslak:
CC: @7sharp9. 
Dave, I think you've mentioned Swift/AppCode (?) having a similar feature, any suggestions about how this should work?

jbeeko:
First off I love it. I waste time fiddling with the mouse trying to get hover up and always having docs for current cursor is great. 

A few comments and half-baked ideas:

1. Since you are forced to the split screen mode rather than a "real" panel you could take advantage of the screen space. Put the information for the cursor portion at the top, the hover information below. 
2. What about treating the information related to the cursor location like intelisense. So if the cursor was on map in  `List.map` the pannel would show other `.` options available?

7sharp9:
I would mainly want lots of detail for a caret point in code, and the option to pin it too.

AlgorithmsAreCool:
I like this. It reminds me of the early demos of Light Table. 

I would love a docked 'panel' that showed signature/docs and possibly even source code for functions in the same project.

Krzysztof-Cieslak:
Some more informations in the panel:
image

Krzysztof-Cieslak:
Demo of the link navigation:

InfoPanel

Krzysztof-Cieslak:
image

MangelMaxime:
This seems really interesting to explore an API and perhaps even documentation from a library? 🎉

@Krzysztof-Cieslak Do you think, it could be a good idea to had syntax highlight support?

Yes, I know it seems like I talk about that often for Ionide  😂 (sorry 🙏)

I don't know if VSCode has an API for that otherwise we could try to use [Code-Lightner](https://github.com/MangelMaxime/Code-Lightner/) for that. 

I didn't update the F# doc yet but it's easy to use from F# too. [Binding source](https://github.com/MangelMaxime/Nacara/blob/55f56710113e7bde32cabed1e6ca1a2278141b1a/src/Libs/CodeLightner.fs)

Krzysztof-Cieslak:
@MangelMaxime, I'd be happy to have syntax highlighting here. Code doesn't have API for that, if we do it we would use your library (BTW, it seems really cool, haven't seen it before). 

The one thing we could additionally consider is creating a separate grammar for the signatures that we could use both here and in tooltips - it's not standard F#, and we have problems with tooltips signature colourization even today.

MangelMaxime:
> @MangelMaxime, I'd be happy to have syntax highlighting here. Code doesn't have API for that, if we do it we would use your library (BTW, it seems really cool, haven't seen it before).

Thank you :)

@Krzysztof-Cieslak Sure, if we have a format that's fixed we can create a dedicated grammar for it and avoid the strange highlights we sometimes get in tooltips.

Krzysztof-Cieslak:
The panel can be locked/unlocked if someone prefers to just navigate with links without updating on cursor move

CC: @kjnilsson

image

Krzysztof-Cieslak:
This is released in `3.36`. Thanks a lot for all additional suggestions - I'll definitely think about them, and I think Info Panel can still be made better.

Issue #1078, opened 2019-04-18
Incorrect and unnecessary error message when creating a new file


isaacabraham:
If you have an existing project or solution open and create a new file (e.g. `CTRL + N`), Ionide pops up immediately with the following warning:

![image](https://user-images.githubusercontent.com/1781813/56368270-602eb180-61f7-11e9-83f0-f15f8b814536.png)

But the file doesn't have a name at this stage (It's just `Untitled-1`) and even if it did - I haven't yet saved it or anything. In this case, I just wanted to create a script - there is absolutely no need for the dialog to appear, and it just serves to distract the developer as well as reduce confidence in Ionide.

isaacabraham:
Whooo :-)

PR #1093, opened 2019-05-07
Remove not in project warning

vbfox:
This change completly remove the "This file is not in any F# project" warning (See #1078)

It's a good warning but doesn't need to be a "in your face" one, so I replaced it with the already existing project indicator in the status bar. It now morph to a message instead of disappearing when the file is not in any project

image

The message is actually also a button that trigger the Forge command to add the current file to the project (That I didn't manage to make work, but debugging or even understanding Forge is for another day)

In addition to the previous implementation the warning also detects untitled (unsaved) files and ignore them.

*PS: Require https://github.com/ionide/ionide-vscode-helpers/pull/28 to be merged / integrated*

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -1480,6 +1480,15 @@
 					"light": "#919191",
 					"highContrast": "#A2E2A2"
 				}
+			},
+			{
+				"id": "fsharp.statusBarWarnings",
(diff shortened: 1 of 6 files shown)


Krzysztof-Cieslak:
Awesome, thanks!

PR #1118, opened 2019-06-06
convert project file paths to uri-style file paths

baronfel:
this matches the contract of the  property and allows for better precision.

This should be lossless: the data we get back from workspacepeek right now is raw file paths, so simple concatenation is fine.


enricosada:
maybe we should choose if we want to move all FSAC path to uri once and forall, to avoid that

PR #1132, opened 2019-06-15
fixed small typo in workbenchViewId function name

tenigma:
As I was reading the source code for another pull request I found this small typo.


tenigma:
I am not sure why some checks are not successful but I guess it cannot be from renaming a function.

PR #1133, opened 2019-06-15
Fix/1130

### LINKED ISSUE #1130: Change "File not in project" status bar item when we're still restoring/loading projects
When you open a project that hasn't been restored or cached it may take a while before we parse all projects.
During this time we can report open filed as "not in project" while in reality we just don't know about it yet - the message is misleading. We should probably change a message a bit while any project is still restoring/loading.

tenigma:
This should resolve #1130.

I just disabled showing "Not in a F# project" in status bar if the workspace is not fully loaded.
While the workspace is loading we do not know if the file is in project or not so we should probably just be quiet.
Is this OK?

### DIFF
--- a/src/Components/QuickInfoProject.fs
+++ b/src/Components/QuickInfoProject.fs
@@ -21,8 +21,9 @@ module QuickInfoProject =
             let proj = Project.tryFindLoadedProjectByFile fileName
             match proj with
             | None ->
-                match te.document with
-                | Document.FSharp when path.extname te.document.fileName <> ".fsx" && not (te.document.isUntitled) ->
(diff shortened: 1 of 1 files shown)


tenigma:
So, I have some troubles testing if this is actually working.
Hi @Krzysztof-Cieslak, when you have some time can you please give me a tip how to test this. Can I somehow slow down loading of projects so that I can see the "Not in a F# project" message in status bar? The loading of projects is so fast I can actually not reproduce this :-). 
Maybe with a workspace with a lot of fsprojs? Do you know a repo with a lot of fsprojs?

PR #1136, opened 2019-06-16
[WIP] Add fake targets outline

matthid:
image


matthid:
This requires https://github.com/ionide/ionide-vscode-helpers/pull/29 (therefore the build is red)

matthid:
(I'd recommend adding those vscode-helpers as git subtree instead of paket dependencies. I actually edited them and didn't realize they are not even under source control)

PR #1137, opened 2019-06-16
Fake target support -> 4 0

matthid:
Replaces https://github.com/ionide/ionide-vscode-fsharp/pull/1136
This depends on https://github.com/fsharp/FsAutoComplete/pull/415 and  ionide/ionide-vscode-helpers#29 

image

(ignore the typo in the screenshot)

image

- [x] Show targets and their dependencies (if possible)
- [x] add support for debugging and running buttons
- [x] Code Navigation via Outline


matthid:
This requires ionide/ionide-vscode-helpers#29 (therefore the build is red)

(I'd recommend adding those vscode-helpers as git subtree instead of paket dependencies. I actually edited them and didn't realize they are not even under source control)

src/Core/LanguageService.fs
```diff
@@ -303,6 +303,45 @@ module LanguageService =
...
+            match client with
+            | None ->
+                Promise.empty
+            | Some cl ->
+                Environment.dotnet
+                |> Promise.bind (fun dotnetRuntime ->
+                    match dotnetRuntime with
+                    | Some r ->
+                        let extensionPath = VSCodeExtension.ionidePluginPath()
+                        let runtimePath = extensionPath + "/.fakeruntime"
```
matthid: I don't think this is a good way to do it. Here is what this is about:

In order to list target's we need to run fake. So this is the folder we download the `fake.dll` portable latest version into.
The following problems arise:
- What if there is a newer version? How can a user ugprade?
- In development, this path is `/release/.fakeruntime` so it potentially gets bundled into the extension? Is this what we want?

The following questions are open:
- Is there a better path to put lazily loaded files into?
- Should we bundle `fake.dll` like we do for `fsautocomplete.dll`? -> I think we don't want to.

src/Core/LanguageService.fs
```diff
@@ -473,7 +512,8 @@ module LanguageService =
         let verbose = "FSharp.verboseLogging" |> Configuration.get false
 
         let spawnNetCore dotnet =
-            let ionidePluginPath = VSCodeExtension.ionidePluginPath () + "/bin_netcore/fsautocomplete.dll"
+            //let ionidePluginPath = VSCodeExtension.ionidePluginPath () + "/bin_netcore/fsautocomplete.dll"
+            let ionidePluginPath = @"C:\proj\ionide-vscode-fsharp\paket-files\github.com\fsharp\FsAutoComplete\src\FsAutoComplete\bin\Debug\netcoreapp2.1\fsautocomplete.dll"
```
matthid: Without this nothing works. This obviously needs to be reverted. But what is the current development story here with lsp?

src/Components/FakeTargetsOutline.fs
```diff
@@ -0,0 +1,287 @@
...
+        )) |> context.subscriptions.Add
+
+        let provider = createProvider ()
+        let treeOptions = createEmpty<CreateTreeViewOptions<Model>>
+        treeOptions.treeDataProvider <- provider
+        treeOptions.showCollapseAll <- Some true
+        let treeView = window.createTreeView(treeViewId, treeOptions)
+        context.subscriptions.Add treeView
+
+        //window.registerTreeDataProvider(treeViewId, createProvider () )
```
matthid: cleanup

src/Core/LanguageService.fs
```diff
@@ -303,6 +303,45 @@ module LanguageService =
...
+                        let runtimePath = extensionPath + "/.fakeruntime"
+                        let req = { TargetRequest.FileName = handleUntitled fn; FakeContext = { DotNetRuntime = r; PortableFakeRuntime = runtimePath }}
+                        cl.sendRequest("fsharp/fakeTargets", req)
+                        |> Promise.map (fun (res: Types.PlainNotification) ->
+                            res.content |> ofJson<Result<Target[]>>
+                        )
+                        |> Promise.onFail (fun o ->
+                            logger.Error("Error in fsharp/fakeTargets request.", o)
+                        )
+                        //|> request<_, Result<Target[]>> "fakeTargets" 0 (makeRequestId())
```
matthid: cleanup

src/Core/LanguageService.fs
Krzysztof-Cieslak: I have patched `build.fsx` that targets my local clone of FSAC. I do changes in FSAC, run `build.cmd LocalRelase` in FSAC repo, and then run `build.cmd Build` in Ionide repo to copy latest built version of FSAC

src/Core/LanguageService.fs
Krzysztof-Cieslak: Maybe this should be per-repository, rather than global? Maybe using `.fake` folder?

src/Core/LanguageService.fs
matthid: Ok, I think we should see if we can improve that, actually not having to restart code was a really nice feature of fsacUrl. But I guess I just have to get used to the new model (just some days after I started liking the old one :D)
So its probably more personal frustration than anything else ;)

src/Core/LanguageService.fs
matthid: Yes I guess you try to address 

> What if there is a newer version? How can a user upgrade? 

In per-repository setting user can `git clean` and then stuff should work again. On the other side when people encounter issues they might take a while to try `git clean`, so I'm not sure this even addresses that (in the sense that it will still cost a lot of people a lot of debugging time)

However, I'm working towards a model where FAKE scripting works well with Ionide. In this "worldview" there is no repository. Maybe I'm just editing single scripts. So now ionide downloads the fake package (~20MB) for each script... 
Additionally, performance might be an issue: When opening a new script Ionide starts downloading. Depending on your current internet speed it will take until everything works. While this is true for packages as well, those are at least in the nuget package cache.

If we decide we can live with those issues I'm strictly against the `.fake` folder. That folder is completely controlled by `fake`. All contents are basically implementation detail and subject to change. So we would use an `.ionide` folder. This also makes people report issues to the correct place in case they realize where their harddrive space is going ;)

matthid:
New feature Debug and Run buttons (with all dependencies and single target mode):

image

matthid:
I think this ready for review, seems to work.

src/Components/FakeTargetsOutline.fs
```diff
@@ -0,0 +1,369 @@
...
+            else
+                if state.Children.ContainsKey key |> not then
+                    let x = {Key = key; Children = new Dictionary<_,_>(); Symbol = symbol}
+                    state.Children.Add(key,x)
+                let item = state.Children.[key]
+                add' item symbol (endIndex + 1)
+
+    let getRoot (doc : TextDocument) =
+        promise {
+            let! o = LanguageService.FakeSupport.targetsInfo doc.fileName
```
matthid: Initially, I thought it is a bad idea to just execute a fake script (with potential side-effects) when opening a script in Ionide.
But Ionide is doing something similar today: You can just hook into msbuild - like paket does - and there you go: arbitrary code execution.

So the only real reason against doing it that way is if we think people have hurting side-effects in their script by accident. But given the design of FAKE such scripts are probably not that useful (and therefore unlikely to exist).

src/Core/LanguageService.fs
matthid: One other solution would be to make this path a configuration option, then we could just configure the Debug path of FSAC similar to fsacUrl. What do you think?

matthid:
Not sure why the build is red, can someone help?

While further testing I noticed the following (which is currently by design). If navigation doesn't work the likely reason is that `Target.initEnvironment()` is missing at the top of your buildscript. I'm not sure if there is a good way to report this scenario to the user.
**Note**: If you click the targets or dependencies on the left, the editor should 'scroll' to the relevant code on the right.

matthid:
Also note that navigation should work no matter which DSL you are using as long as it is backed by `Fake.Core.Target` under the hood. For example https://github.com/vbfox/FoxSharp/blob/master/src/BlackFox.Fake.BuildTask/Readme.md should work as well /cc @vbfox

matthid:
I think navigation into targets defined in a `#load`ed script might not work, but we can fix that later.

src/Components/FakeTargetsOutline.fs
```diff
@@ -0,0 +1,366 @@
...
+                        { name = "Fake Script Debugging"
+                          ``type`` = "coreclr"
+                          request = "launch"
+                          preLaunchTask = None
+                          program = fakeRuntime
+                          args = args
+                          cwd = scriptDir
+                          console = "externalTerminal"
+                          stopAtEntry = false
+                          justMyCode = false
```
matthid: I'm not sure why we need to deactivate this or what "my Code" means here. If I set this to `true` no breakpoint hits.
How can I tell the debugger that the current file is "my code"? Anyone understands the details here?
The problem with `false` is that you can see all the stack from the FAKE runtime and potentially unrelated stuff...

matthid: Can someone please ping someone who knows something?

Krzysztof-Cieslak: I believe this is exactly the same as Just My Code in VS - https://docs.microsoft.com/en-us/visualstudio/debugger/just-my-code?view=vs-2019

src/Core/LanguageService.fs
Krzysztof-Cieslak: > One other solution would be to make this path a configuration option, then we could just configure the Debug path of FSAC similar to fsacUrl. What do you think?

Yes, I think this would be good

src/Components/FakeTargetsOutline.fs
```diff
@@ -0,0 +1,366 @@
...
+                          requireExactSource = false }
+                    if doDebug then
+                        let! res = debug.startDebugging(JS.undefined, unbox cfg)
+                        ()
+                    else
+                        let! dotnet = Environment.dotnet
+                        match dotnet with
+                        | None ->
+                            let! _ = window.showErrorMessage("Cannot start debugging as no dotnet runtime was found. Consider configuring one in ionide settings.")
```
Krzysztof-Cieslak: This message is for running case not debugging. `"Cannot start FAKE script... "`

src/Components/FakeTargetsOutline.fs
matthid: > The .NET Framework debugger considers optimized binaries and non-loaded .pdb files to be non-user code.
> Three compiler attributes also affect what the .NET debugger considers to be user code:
>
>    DebuggerNonUserCodeAttribute tells the debugger that the code it's applied to isn't user code.
>    DebuggerHiddenAttribute hides the code from the debugger, even if Just My Code is turned off.
>     DebuggerStepThroughAttribute tells the debugger to step through the code it's applied to, rather than step into the code.
>
> The .NET Framework debugger considers all other code to be user code.

Ok none of these attributes was set but I saw in the debug output:

`Not loading <compiledDll_hash>.pdb because just my code is enabled`.
So the debugger is not loading the pdb because just my code is enabled and it considers the dll not 'my code' because no pdb is loaded? Circle reasoning? I'm sure the dll has `--optimize-` enabled

src/Core/LanguageService.fs
Krzysztof-Cieslak: Let's handle it in other PR

matthid: Ok Thanks

src/Components/FakeTargetsOutline.fs
matthid: Yes indeed, nice catch!

matthid:
Ok it now looks like this:
image

I took the icons from https://github.com/microsoft/vscode/tree/master/src/vs/base/browser/ui/dialog, not sure if this OK, probably not. Ideas?

matthid:
Ideally, we would add Icons for "soft" and "hard" dependencies as well...

src/Components/FakeTargetsOutline.fs
matthid: Should we ignore this for now and note as "known limitation"? To be honest, I have no idea what to do. 
Just unfortunate that the experience is a bit unfortunate... At least it works.

src/Components/FakeTargetsOutline.fs
```diff
@@ -0,0 +1,402 @@
...
+            match m.Declaration with
+            | Some decl -> 
+                let line = decl.Line
+                let args =
+                    createObj [
+                        "lineNumber" ==> line - 1
+                        "at" ==> "center"
+                    ]
+
+                vscode.commands.executeCommand("revealLine", args)
```
matthid: Is there another way to do this, this is the code from the previous F# outline but sometimes it feels broken and doesn't work well?

We can improve that later on.

Issue #1012, opened 2019-02-07
Can the format of the auto-generated doc comments be updated?


baronfel:
I noticed while working on doc generation in [MiniScaffold](https://github.com/TheAngryByrd/MiniScaffold/pull/110) that the markdown rendered by FSharp.Formatting is a little rough when the sections are delimited by `*` instead of `#` headings. When `#` are provided the parser gives a lot more detail to the structure that we can use to render sections a bit better, so I'd like to see if you are open to changing the current auto-generated comment headings.

Let me give an example of the structure we get back from FSharp.Formatting in different scenarios. Here is our sample code:

```fsharp
module Say =

    /// **Person**
    type Person = {
        /// First
        First : string
        Last : string
        FavoriteNumber : int
        DateOfBirth : DateTimeOffset
    }


    /// **Description**
    ///
    /// **Parameters**
    ///   * `person` - parameter of type `Person`
    ///
    /// **Output Type**
    ///   * `string`
    ///
    /// **Exceptions**
    ///
    let helloPerson (person : Person) =
        sprintf
            "Hello %s %s. You were born on %s and your favorite number is %d."
            person.First
            person.Last
            (person.DateOfBirth.ToString("o"))
            person.FavoriteNumber

    /// #### Description
    /// I do nothing, ever.
    ///
    /// #### Parameters
    ///   * `name` - parameter of type `'a`
    ///
    /// #### Output Type
    ///   * `unit`
    ///
    /// #### Exceptions
    /// None
    let nothing name =
        name |> ignore


    /// **Description**
    ///
    /// **Parameters**
    ///   * `name` - parameter of type `string`
    ///
    /// **Output Type**
    ///   * `string`
    ///
    /// **Exceptions**
    ///
    [<CompiledName("Hiya")>]
    let hello name =
        sprintf "Hello %s" name
```

and here is the matching Comment structure for the `helloPerson` comment block:
```
{Blurb =
  "<p><strong>Description</strong></p>
<p><strong>Parameters</strong>
* <code>person</code> - parameter of type <a href="mylib-_1-say-person.html" title="Person"><code>Person</code></a></p>
<p><strong>Output Type</strong>
* <code>string</code></p>
<p><strong>Exceptions</strong></p>

";
 FullText =
  "<p><strong>Description</strong></p>
<p><strong>Parameters</strong>
* <code>person</code> - parameter of type <a href="mylib-_1-say-person.html" title="Person"><code>Person</code></a></p>
[...]

PR #1175, opened 2019-07-11
Convert doc comment to XML

MangelMaxime:
Related #1012

Instead of Markdown, we now use XML format:

```fs
    /// <summary>
    /// 
    /// </summary>
    /// <param name="url"></param>
    /// <param name="data"></param>
    /// <param name="decoder"></param>
    /// <param name="properties"></param>
    /// <typeparam name="Response"></typeparam>
    /// <returns></returns>
    static member tryDelete<'Response>(url : string,
                                       data : JsonValue,
                                       decoder : Decoder<'Response>,
                                       ?properties : RequestProperties list) =
```

Rules for the generation:

1. Always generate the `summary` section
2. Generate the list of `param` if any
3. Generates the list of `typeparam` (generic info) if any
4. Always generate the `returns` section

------------

@Krzysztof-Cieslak Because we were supporting markdown, we have the following highlights:

Capture d’écran 2019-07-11 à 16 13 36

Are you ok to remove `markdown` support for inlined comment starting by `///` and replace it with `xml` support?

Capture d’écran 2019-07-11 à 16 13 51

We can't support both, I tried to set up support for both but it's the same result as if I was adding only `markdown` support.


MangelMaxime:
CI seems to fail on VSCE installation but I didn't touch that part.

MangelMaxime:
About the change for highlights, I will probably improve the grammar extension in order to provide a better experience than raw XML. Will send a POC to Ionide fsgrammar repo and discuss it their.

Krzysztof-Cieslak:
This is good, but it makes me sad anyway. Markdown was so much better idea, I'm sad no one else wanted to support it.

> Are you ok to remove markdown support for inlined comment starting by /// and replace it with xml support?

Yes, that sounds reasonable.

MangelMaxime:
> This is good, but it makes me sad anyway. Markdown was so much better idea, I'm sad no one else wanted to support it.

Makes me sad and I was a big fan of Markdown comments because they had a good look in the source code and the tooltip but only in Ionide...

If someone is using VS or Rider it would have looked really strange...

You can merge this PR when you feel ready for it :)

baronfel:
How are we looking here, anything stopping this from getting built? Does the FSAC tooltip work overlap with this at all?

MangelMaxime:
Oops, I need to check.

I totally forget about this PR and added another version of the XML generation in #1259.

Let me check which version should be used.

MangelMaxime:
Close in favour of #1259 

It makes sense to update the generation and the parser at the same time.

PR #1184, opened 2019-07-17
WIP: Upgrade to latest fable.

baronfel:
Not latest library bindings, but latest versions of the compiler and underlying tooling.

This fixes the odd project cracking issue we saw where the more recent features that required 'activation' didn't compile.

Haven't tested locally much, about to do that now.


baronfel:
Definitely broken locally, there's something odd happening with closure compilation.  I added some missing dependencies based on reading some sample webpack configs, still digging though.

baronfel:
Current root problems seems to be around compiling literals.

Example F# code:
```fsharp
    let private programFilesX86 =
        let wow64 = Globals.``process``.env?``PROCESSOR_ARCHITEW6432`` |> unbox<string>
        let globalArch = Globals.``process``.env?``PROCESSOR_ARCHITECTURE`` |> unbox<string>
        let envValue =
            match wow64, globalArch with
            | "AMD64", "AMD64" | null, "AMD64" | "x86", "AMD64" -> Globals.``process``.env?``ProgramFiles(x86)``
            | _ -> Globals.``process``.env?``ProgramFiles``
        let unboxed = unbox envValue
        if unboxed = null then @"C:\Program Files (x86)\"
        else unboxed
```

compiles to 

```js
const programFilesX86 = (() => {
  var clo2, clo2$$1;
  const wow64 = (clo2 = process.env.PROCESSOR_ARCHITEW6432, function (arg20) {
    return clo2(arg20);
  });
  const globalArch = (clo2$$1 = process.env.PROCESSOR_ARCHITECTURE, function (arg20$$1) {
    return clo2$$1(arg20$$1);
  });
  let envValue;
  const matchValue$$1 = [wow64, globalArch];
  var $target$$3;

  if (matchValue$$1[0] === "AMD64") {
    if (matchValue$$1[1] === "AMD64") {
      $target$$3 = 0;
    } else {
      $target$$3 = 1;
    }
  } else if (matchValue$$1[0] === "x86") {
    if (matchValue$$1[0] === null) {
      if (matchValue$$1[1] === "AMD64") {
        $target$$3 = 0;
      } else {
        $target$$3 = 1;
      }
    } else if (matchValue$$1[1] === "AMD64") {
      $target$$3 = 0;
    } else {
      $target$$3 = 1;
    }
  } else if (matchValue$$1[0] === null) {
    if (matchValue$$1[1] === "AMD64") {
      $target$$3 = 0;
    } else {
      $target$$3 = 1;
    }
  } else {
    $target$$3 = 1;
  }

  switch ($target$$3) {
    case 0:
      {
        const clo2$$2 = process.env["ProgramFiles(x86)"];

        envValue = function (arg20$$2) {
          return clo2$$2(arg20$$2);
        };

        break;
      }

    case 1:
      {
        const clo2$$3 = process.env.ProgramFiles;

        envValue = function (arg20$$3) {
          return clo2$$3(arg20$$3);
        };

        break;
      }
  }

  const unboxed = envValue;
  return unboxed === null ? "C:\\Program Files (x86)\\" : unboxed;
})();
```

but that value isn't a string (and isn't even a function that can be called!)
```
> programFilesX86

function (arg20$$3) { … }
  [[FunctionLocation]]:internal#location
  [[Scopes]]:Scopes[7]
    0:Block (./src/Core/Environment.fs.programFilesX86) {type: "block", name: "./src/Core/Environment.fs.programFilesX86", object: Object}
    1:Closure (./src/Core/Environment.fs.programFilesX86) {type: "closure", name: "./src/Core/Environment.fs.programFilesX86", object: Object}
    2:Closure {type: "closure", name: "", object: Object}
    3:Closure (./src/Core/Environment.fs) {type: "closure", name: "./src/Core/Environment.fs", object: Object}
    4:Closure {type: "closure", name: "", object: Object}
    5:Closure {type: "closure", name: "", object: Object}
    6:Global {type: "global", name: "", object: global}
  arguments:TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
  caller:TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
  length:1
  name:"envValue"
  prototype:Object {constructor: }
  __proto__:function () { … }

> programFilesX86()
TypeError: clo2$$3 is not a function
```

This is bad because we expect this to be bound to a string (ie we pass this value directly into string.replace)

Krzysztof-Cieslak:
CC: @MangelMaxime

baronfel:
My first guess is that the ? operator works differently now? I'm reading in some places that if you want to 'apply' it directly you need to wrap the accessed element in parens: Globals.``process``.env?(``PROCESSOR_ARCHITEW6432``)

MangelMaxime:
I can't yet take a look at the project because I can't build FSAC when referenced from Ionide while I can build FSAC when cloned from the repo by myself.

But when I tried to adapt the code you provided for a browser

Using something like:

```fs
let wow64 = Browser.Dom.window.document?``PROCESSOR_ARCHITEW6432`` |> unbox<string>
```

I get the expected behaviour:

```js
  var wow64 = window.PROCESSOR_ARCHITEW6432;
  var globalArch = window.PROCESSOR_ARCHITECTURE;
```

I will try to fix my env and take a look at the issue.

> I'm reading in some places that if you want to 'apply' it directly you need to wrap the accessed element in parens

It was the case when you wanted to call the function or for accessing via `myobj.[some_value]`.

.vscode/settings.json
```diff
@@ -17,6 +17,9 @@
     ".git",
     "paket-files",
     "release"
-  ]
+  ],
```
MangelMaxime: I don't think we want to add this setting for everyone, do we?

MangelMaxime:
> (and isn't even a function that can be called!)

This is normal that you can't call the function because it's a value :) 

Look at the  JavaScript code:

```js
const programFilesX86 = (() => {
// ...
})();
```

`programFilesX86`  create a function that is directly call. It’s an Immediately-Invoked Function Expression, or IIFE for short. It executes immediately after it’s created.

Because in F# everything is a function even values and also it helps create scope/closure.

MangelMaxime:
If you use this version of the code:

```fs
let private mapToString (value : obj) =
    if isNull value then
        Some (unbox<string> value)
    else
        None

let private programFilesX86 =
    let wow64 = Globals.``process``.env?``PROCESSOR_ARCHITEW6432`` |> mapToString
    let globalArch = Globals.``process``.env?``PROCESSOR_ARCHITECTURE`` |> mapToString
    match wow64, globalArch with
    | Some "AMD64", Some "AMD64"
    | None, Some "AMD64"
    | Some "x86", Some "AMD64" -> Globals.``process``.env?``ProgramFiles(x86)`` |> mapToString
    | _ -> Globals.``process``.env?``ProgramFiles`` |> mapToString
    |> function
        | Some detected -> detected
        | None -> @"C:\Program Files (x86)\"
```

Then you get a string value as a result.

I am not sure if this is a bug in Fable implementation or not. But what is happeding in your implementation is that you have checkes like `matchValue$$1[0] === null` but `matchValue$$1[0]` is `undefined`. And it seems to break the condition.

By using `isNull` we are using `equal` implementation coming from Fable and it's working at expected I think.

I also checked that you can applu `string.replace` on it, on my computer I got this result:

```fs
JS.console.log programFilesX86
// C:\Program Files (x86)\

JS.console.log (programFilesX86.Replace("x86", "maxime hey hey"))
// C:\Program Files (maxime hey hey)\
```

PS: The code in this PR doesn't compile for me. Because, the Helpers modules has not been updated to latest version of `Promise` signatures.

Example:

```
/Users/maximemangel/Workspaces/Github/ionide/ionide-vscode-fsharp/paket-files/ionide/ionide-vscode-helpers/src/Helpers.fs(93,19): (93,110) error FSHARP: This expression was expected to have type
    'obj -> 'T'    
but here has type
    'Func<obj,U2<'T,PromiseLike<'T>>>'     (code 1)
```

If you have any more question feel free to ask :)

baronfel:
Thanks for diving in @MangelMaxime! I'll try to take what you've given me here and finish this up soon.

baronfel:
Checkpointing: Maxime's sample did get me past the initial startup failure, we're now at least getting to the spawning of FSAC. Not getting any `projectLoad` events firing from the `initialize` handler so that's perplexing.

MangelMaxime:
@baronfel How are you testing the application? 

I am asking because last time when I tried to execute your branch, I wasn't able to compile the project because the Ionide Helpers were not updated. So there was some Fable compiler error reported.

If you still have not fixed the helpers (or have Fable compiler errors) it's probably better to have everything compiling because I am not sure of what is being generated.

baronfel:
I'm testing via the F5 build-and-run-new-extension-host mechanism that's set up. I don't have the same build failure you seem to with the ionide-vscode-helpers/Helpers.fs file, which is confusing to me because that should be locked down by the paket github file ref.  I'll try clearing out my local versions and restoring again.

baronfel:
Yes that was it. I'll zero in on fixing those then.

Krzysztof-Cieslak:
What's the status of this one?

baronfel:
Still stalled, I've not made any progress since the last commits.

7sharp9:
In terms of publishing a nuget with latest vscode bindings, is that easier to get nailed first?

PR #1200, opened 2019-08-20
Fix - add content security policy to info panel web view

### LINKED ISSUE #1199: Webview does not set a content security policy
Hi, I'm the developer of VS Code's webview API. I noticed that your extension seems to create a webview that does not set a [content security policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP). All webviews (even very simple ones) should set a content security policy. This helps limit the potential impact of content injections and is generally a good measure for defense in depth. 

We've documented how to add a content security policy to VS Code webviews [here](https://code.visualstudio.com/api/extension-guides/webview#content-security-policy). Please  add the most restrictive content security policy possible to your webview. I am not aware of any immediate security issues with your extension but having a restrictive content security policy is important to help protect users of your extension.

---

*Also note that in development mode, in VS Code 1.38 you should also see a warning if you create a webview that does not set a content security policy: https://github.com/microsoft/vscode/issues/79248*

ZaymonFC:
Attempting to resolve #1199.

Currently unsure if `content="default-src 'none'` is the correct security policy to use.

https://code.visualstudio.com/api/extension-guides/webview#security

### DIFF
--- a/src/Components/InfoPanel.fs
+++ b/src/Components/InfoPanel.fs
@@ -39,6 +39,7 @@ module InfoPanel =
                     sprintf """
                     <html>
                     <head>
+                    <meta http-equiv="Content-Security-Policy" content="default-src 'none';">
                     <style>
                     pre {color: var(--textCodeBlock.background)}
                     </style>


Krzysztof-Cieslak:
Thanks a lot!

PR #1210, opened 2019-09-07
Fix off-by-one bug in LineLens

### LINKED ISSUE #1195: Different type signatures for identical functions.
**Describe the bug**
Different type signatures for identical functions.

**To Reproduce**
```fsharp
let a x y z = x (y z)
let aa x y z = x (y z)
```

You can see the inferred type signature for the first function is incorrect (`'a -> 'b`) in the screenshot.

**Expected behaviour**
Both functions should have the signature:
```
('a -> 'b) -> ('c -> 'a) -> c' -> b'
```

**Screenshots**
image

 - OS: Windows and OSX
 - Ionide version: 4.0.6
 - VSCode version: July 2019
 - dotnet SDK version: 2.2.204
 - mono / .Net Framework version: 5.18.1.3

vbfox:
Fixes #1195

image

### DIFF
--- a/src/Components/LineLens.fs
+++ b/src/Components/LineLens.fs
@@ -193,7 +193,7 @@ module DecorationUpdate =
                 LanguageService.signatureData
                     fileName
                     range.StartLine
-                    range.StartColumn
+                    (range.StartColumn - 1)
             let signaturesResult = if isNotNull signaturesResult then Some signaturesResult else None
             return signaturesResult |> Option.map (fun r -> range, formatSignature r.Data)
         }


vbfox:
I find very weird to have to do that and suspect that the real off by one is in FSAC as we should be able to ask for symbols and send back the same ranges to FSAC without adjustment but 🤷‍♂️

Krzysztof-Cieslak:
Have you checked if normal CodeLenses behavior is correct? If not we definitely should fix that in FSAC

ZaymonFC:
🙌 Very nice.

PR #1241, opened 2019-10-10
Set/pick up configured dotnet root

baronfel:
Needs to be tested still but this is the grunt work to plumb through the pluggable dotnet root from FSAC.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,7 +2,7 @@ name: $(Rev:r)
 jobs:
 - job: IonideFsharp
   pool:
-    vmImage: 'vs2017-win2016'
+    vmImage: 'windows-2019'
   steps:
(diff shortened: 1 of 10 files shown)


baronfel:
*sigh* and now AZDO 2019 builds don't 2.1.801 either.

Krzysztof-Cieslak:
So, it seems that right now we have `dotnetLocation` and `dotNetRoot` settings, what's the deal with it?

Krzysztof-Cieslak:
Especially that `dotnetLocation` is not declared in package.json, and it's mentioned only in the error message on FSAC startup.

Krzysztof-Cieslak:
And don't worry about experimental build - I'll remove it for now, we don't update experimental extension at the moment.

Issue #1242, opened 2019-10-11
Terminal is slow when sending content to FSI


ronaldschlenker:
I'm not sure if that's a bug, anyway: When I send some text to FSI, the terminal is very slow. It's like piping text via modem to your C64 :) Anyone else having that impression? I could make a gif or video to show you what I meen if necessary.

ronaldschlenker:
Of course.

For example, I have a powershell file, containing 1600 lines of this: ```$ded = "sdfsdfsdfg"```

Sending all 1600 lines (by marking all -> F8) to the powershell terminal, it is done in under 1s.

Sending 1600 lines of ```let ded = "sdfsdfsdfg``` to FSI, it takes 26s. I'm pretty sure that the time is not taken by the evaluation step of FSI, but just by sending the chars.

Here are some infos:

* tested with Ionide 4.1.0 (F# 4.5) and 4.2.0 (F# 4.7) - all oldschool .Net FSI
* fsiAnyCpu is used
* Windows 10 (latest updates)
* VSCode Canvas terminal rendering is forced ("terminal.integrated.rendererType": "canvas")

When I monitor the processes in taskmgr, the fsiAnyCpu, VsCode and a conhost process take each around 15 - 20% CPU (2 physical cores + 2 HT).

VSCode:
Version: 1.39.1 (user setup)
Commit: 88f15d17dca836346e787762685a40bb5cce75a8
Date: 2019-10-10T23:31:28.683Z
Electron: 4.2.10
Chrome: 69.0.3497.128
Node.js: 10.11.0
V8: 6.9.427.31-electron.0
OS: Windows_NT x64 10.0.18362

I hope this helps. If you need something more, let me know.

baronfel:
Thanks, that helps a lot. One final question: when you say

> Sending all 1600 lines (by marking all -> F8) to the powershell terminal

Do you mean you're running fsi inside the powershell terminal? Or do you mean that you're using the powershell terminal to just paste text into to act as a baseline?

ronaldschlenker:
> Or do you mean that you're using the powershell terminal to just paste text into to act as a baseline?

Exactly; that was the intention. I wanted to show that it might really be an issue in the FSI terminal of Ionide and it's not a basic issue of all terminals in VSCode.

baronfel:
[...]

PR #1243, opened 2019-10-11
no longer chunk messages because underlying node lib is better

baronfel:
Per https://github.com/ionide/ionide-vscode-fsharp/issues/1242#issuecomment-541135786 the underlying node library is better so we may not need to chunk the message anymore.


baronfel:
The linked issue shows an example sending almost 3000 characters at once and that working, so this seems safe.

PR #1248, opened 2019-10-29
fix FSI spawning logic when the user is .net sdk but lacks .net framework

### LINKED ISSUE #1247: FSI on Windows, dotnet core only does not work
**Describe the bug**

On a Windows machine with *only* dotnet core 3 SDK installed, sending code from an `fsx` file to FSI fails with `Failed to spawn FSI, please ensure it's in PATH` + `Failed to send text to FSI`.

**To Reproduce**

Clean Windows 10 machine, with dotnet core 3, VS Code 1.39.2 and Ionide 4.2.0.
Create an `.fsx` file, write some valid F# code, select and send to FSI with `Alt+Enter`.

**Expected behaviour**

The code should be sent to FSI for evaluation.

**Additional context**

The problem arises if *ONLY* dotnet core is installed, and not the full framework.

baronfel:
When users are on windows and using .Net Core, but _do not_ have .Net Framework installed, any attempt to launch an FSI shell will crash.

This is because in this situation:
* FSAC returns null for FSI path (because it is looking in the older framework paths)
* Ionide probes in framework paths for it (and is not found)
* and we unconditionally look for FSI regardless of the .Net Core opt-in a user may have done.

So instead this PR breaks out the 'support matrix' into four cases:
* .net core, 'dotnet' binary found -> invoke `dotnet fsi`
* .net core, 'dotnet' binary not found -> fail with message
* .net framework, 'fsi' path found -> invoke (mono) 'fsi'
* .net framework, 'fsi' path not found -> fail with message

This should fix #1247.

### DIFF
--- a/src/Components/Fsi.fs
+++ b/src/Components/Fsi.fs
@@ -62,13 +62,15 @@ module Fsi =
         promise {
             if isSdk
             then
-                match! LanguageService.dotnet() with
+                let! dotnet = LanguageService.dotnet ()
+                match dotnet with
                 | Some dotnet ->
                     return dotnet, [|yield "fsi"; yield! parms |]
                 | None ->
                     return failwith "dotnet fsi requested but no dotnet SDK was found."
             else
-                match! LanguageService.fsi () with
+                let! fsi = LanguageService.fsi ()
+                match fsi with
                 | Some fsi ->
                     return fsi, parms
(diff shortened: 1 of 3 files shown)


baronfel:
The ScriptRunner component does need to be updated in a similar fashion.

PR #1288, opened 2020-01-08
Update to latest FsAutoComplete

nojaf:


nojaf:
Thanks for the quick merge @Krzysztof-Cieslak!
Any idea when the next Ionide release would be?
Of perhaps an insiders release if that is still a thing.
Would like to gather some feedback from that configuration file.

PR #1298, opened 2020-02-03
Update to latest FSAutocomplete

nojaf:
Contains update to [Fantomas 3.2.0](https://github.com/fsharp/FsAutoComplete/commit/1f41fcde85af1f6945e082cfce8d919d8a0c869a).

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -362,7 +362,8 @@ module ExperimentalExtension =
         run "git" "checkout vnext" fsacDirPath
 
         // local release it
-        run (if isUnix then "build.sh" else "build.cmd") "LocalRelease" fsacDirPath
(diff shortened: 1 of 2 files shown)


nojaf:
Some other changes are needed for the build pipelines to work.
Things like
```
  steps:
  - task: UseDotNet@2
    displayName: 'Use exact .NET Core sdk'
    inputs:
      packageType: sdk
      version: 2.1.801
```
won't work for building FSAC. I'm not familiar enough with this setup to fix this I'm afraid.

baronfel:
yes, it looks like we need to move from paket-git-includes of FSAC to an actual artifact. @Krzysztof-Cieslak looks like we need to start tagging/building releases of FSAC again.

PR #1327, opened 2020-03-22
Fsi watcher

Krzysztof-Cieslak:


forki:
What is it doing?

baronfel:
@forki it's a viewer for let-bound values in an FSI session, IIRC

forki:
Thx. Awesome

Chet Husk <notifications@github.com> schrieb am So., 22. März 2020, 16:53:

PR #1329, opened 2020-03-24
Recognize .gitignore paths that contain slashes at the start or the end.

teo-tsirpanis:
Some might ignore the directory `.ionide/`, but the extension would still recommend `.ionide` to be ignored.


Krzysztof-Cieslak:
Thanks a lot!

PR #1358, opened 2020-05-03
Fix lines split for Windows system

Viridovics:
Hello!

Every times when I open F# project, I recieve Ionide message.

image

I think this fix may be helpful.


Krzysztof-Cieslak:
Thanks!

PR #1379, opened 2020-06-08
No prompt to update gitignore when there is none

### LINKED ISSUE #1336: Asks to add Ionide entries to .gitignore when there is no .gitignore.
When opening a folder which is not a git repository and has no .gitignore file, it still offers to add Ionide entries to the file and will create a empty .gitignore to do so.

Create an empty folder.
Add a .fsx script file.
Open the folder with VS Code
A pop up appears asking to add missing entries to the .gitignore file.
Select Add Entries
A .gitignore file is created with just the two Ionide entries.

**Expected behaviour**
No pop up should occur if there is no existing .gitignore and / or the folder is not a git repo.

 - OS: Windows 10 1909 18363.752 
 - Ionide version: 4.8.1
 - VSCode version: 1.43.2
 - dotnet SDK version: 3.1.300-preview-015048

josefblaha:
Fixes #1336. The extension no longer offers to add its patterns to .gitignore file when the file does not exist.

### DIFF
--- a/src/Components/Gitignore.fs
+++ b/src/Components/Gitignore.fs
@@ -10,9 +10,13 @@ module Gitignore =
     open Ionide.VSCode.FSharp
 
     let GITIGNORE_KEY = "FSharp.suggestGitignore"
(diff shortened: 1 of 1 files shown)


Krzysztof-Cieslak:
Awesome, thanks a lot!

Issue #1380, opened 2020-06-09
Project from fable-getting-started won't load


aspnetde:
I followed the [steps provided](https://zaid-ajaj.github.io/the-elmish-book/#/chapters/fable/hello-world) by Zaid in his Elmish Handbook to set up a simple hello world application.

The following works:

```bash
git clone https://github.com/Zaid-Ajaj/fable-getting-started.git
cd fable-getting-started
npm install
npm run build
```

Fable compiles and loading `./fable-getting-started` in VSCode with Ionide 4.11.1 works.

However, npm is reporting 17 vulnerabilities right now when running `npm install`. When I run `npm audit fix`, it claims that 14 of those had been fixed. Fable would still compile, but Ionide does not load the project anymore.

<img width="690" alt="image" src="https://user-images.githubusercontent.com/1322584/84142511-35d03100-aa55-11ea-84db-c01bc155806e.png">

I can't say for sure it's related to the npm audit call, but that's how I could reproduce it.

--

I am on macOS 10.15.4, running netcore 3.1.300 and node 14.4.0.

forki:
I think the repro is not related to npm. Here is how i can repro the ionide issue:

1) `git clone https://github.com/Zaid-Ajaj/fable-getting-started.git`
2) `cd fable-getting-started`
3) `code .`
4) navigate to `App.fs` and see that syntactical highlighting actually kicks in
5) look at `src/obj' and see that `project.assets.json` was created. this implies ionide has properly run  `dotnet restore` internally
6 ) add `open Browser.Dom` in `App.fs`
7) notice red squigglies. this implies ionide did not reload the project after `project.assets.json` was created
8) close VS Code
9) `code .`
10) navigate to `App.fs` and notice that red squigglies are gone and `open Browser.Dom` is proeprly detected as "unused"

aspnetde:
Correct.

forki:
@Krzysztof-Cieslak same happening when I add sln in root

isaacabraham:
I think that this is a dup of #1377

forki:
yes but with repro ;-)

Krzysztof-Cieslak:
Should be fixed in `4.12.1`

forki:
Awesome. Will test tomorrow morning. Thanks a lot

Krzysztof Cieślak <notifications@github.com> schrieb am Do., 11. Juni 2020,
15:04:

aspnetde:
[...]

PR #1382, opened 2020-06-09
Language service is returning error

forki:
related #1380 

This won't solve #1380 but it properly propagates the error from fsauto into ionide

### DIFF
--- a/src/Core/LanguageService.fs
+++ b/src/Core/LanguageService.fs
@@ -296,9 +296,12 @@ module LanguageService =
             cl.sendRequest("fsharp/project", req)
             |> Promise.map (fun (res: Types.PlainNotification) ->
                 let res = res.content |> ofJson<obj>
-                deserializeProjectResult (res |> unbox)
+                match res?Kind |> unbox with
+                | "error" ->
+                    res?Data |> parseError |> Error
+                | _ ->
+                    deserializeProjectResult (res |> unbox) |> Ok
             )
-
 
     let workspacePeek dir deep excludedDirs =
         let rec mapItem (f : WorkspacePeekFoundSolutionItem) : WorkspacePeekFoundSolutionItem option =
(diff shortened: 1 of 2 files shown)


forki:
image

forki:
as you can see the status changed. it's no longer "not restored", but now it's "load failed".

But fsauto returns: Code:102 Message: this project was not loaded due to some internal error.
 whatever that means!?

forki:
@Krzysztof-Cieslak is that build error related? doesn't seem so

forki:
ok one level down the rabbit hole. Maybe this ionide change is just not the right place. I found: 

image

is this correct? Or should this be some kind of error!?

forki:
next step. I followed the trail of that error. I think it's raised inside of 

image

so that means FCS!? Fuck me

baronfel:
That GetProjectOptions is in Dotnet-Proj-Info, so you're safe for a bit :)

forki:
oh I don't mind working in FCS. I meant: that rabbit hole is fucking deep

Krzysztof-Cieslak:
Maybe we could try generating binlog and checking what it says - `FSharp.generateBinlog` setting

Krzysztof-Cieslak:
I'll merge this at it fixes error propagation bug - loading problem needs to be fixed in FSAC/dotnet-proj-info so it's not on Ionide side anyway.

forki:
I wonder if we run in some caching issue here. If I run dotnet proj fsc-args on my repo folder then dotnet-proj-info correctly restores and the prints the fsx params. But what if something between all of this is caching the unloaded state!?

isaacabraham:
This issue hasn't always been around though (in my experience) - I only observed it in the past month or so, maybe less.

isaacabraham:
On a new Farmer project (using Farmer.Template), I can repro this as well. Might not be helpful, dunno, but in the status bar I see: "Not in a F# project (Still loading)". When I mouse over, I get: "Program.fs is not in any project known to Ionide". Meanwhile next to it it shows "Project loading..." for the fsproj and never returns.

PR #1400, opened 2020-07-12
Update FSAC

nojaf:


baronfel:
@Krzysztof-Cieslak any idea what's happening here? the vsce tool can't be found, and I guess that's required to actually build the package?

baronfel:
Thanks @nojaf for the kick in the pants :)

Issue #1041, opened 2019-03-05
ionide creates an .ionide directory when not loading F# projects or files


TheGoddessInari:
### Problem

If you open a folder, an `.ionide` directory is required for the symbol cache, even if there's no F# files, no project files, no intent to use F# , and hence no symbols to save.
I'm not certain what triggers this, I'm seeing this on some things but not others. I always get an `.ionide` folder with https://github.com/servo/servo.

### Steps to reproduce

1. Open VS Code
2. Open a folder.
3. An `.ionide` folder is created because of symbol cache.

### Machine infos

* Operating system: **Windows_NT**
* Arch: **x64**
* VSCode: **1.31.1**
* Runtime: **netcore**
* Dotnet version: **3.0.100-preview3-010426**
* MSBuild version:
```shell
Microsoft (R) Build Engine version 16.0.443+g5775d0d6bb for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.
```

TheGoddessInari:
Looks like it's related to background caching unconditionally invoking FsAutoComplete. 

Examining FsAutoComplete, it seems to create the directory unconditionally when starting its persistent state, so probably the error is possibly invoking FsAutoComplete in situations with no F# projects or data.

Source code probably needs to be combed over to ensure that nothing will start up FsAutoComplete inappropriately or unconditionally.

Krzysztof-Cieslak:
Can't reproduce on latest Ionide and VSCode.

rpus:
I've (very recently, since you closed this, and, I think, for the first time) started getting this unwanted behaviour too, seemingly on all my projects, none of which have anything to do with F#. I'm on VS Code 1.35.1 and Ionide-fsharp 4.0.0, so I'm guessing the latter has (re-)introduced this behaviour. Hope that helps...

jaredswarts55:
This is happening to me as well, I wonder if this should be reopened.

fogzot:
Happening here too. Ionide 4.0.5 on VS Code 1.36.1.

Adohk:
Same here I've had to disable ionide when not using F#

vbfox:
I've seen it quite a few times in projects without any F# or .Net but I don't have a good repro, need ot investigate.

There's clearly something fishy.

mairaw:
I also get this all the time when working on .NET documentation such as https://github.com/dotnet/docs

Krzysztof-Cieslak:
[...]

PR #1401, opened 2020-07-13
Remove .sln from activation event

slang25:
Removing `.sln` from list of activation events, this is primarily due to this related issue:
https://github.com/fsharp/FsAutoComplete/issues/620

Thinking about it the combination of other events should be sufficient. What are your thoughts?

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -1629,7 +1629,6 @@
 		"workspaceContains:**/*.fs",
 		"workspaceContains:**/*.fsx",
 		"workspaceContains:**/*.fsproj",
-		"workspaceContains:**/*.sln",
 		"onLanguage:fsharp",
 		"onCommand:fsi.Start",
 		"onCommand:fsharp.NewProject"


Krzysztof-Cieslak:
As described before (https://github.com/ionide/ionide-vscode-fsharp/issues/1041#issuecomment-576668861) this is by design behaviour.

slang25:
Thanks @Krzysztof-Cieslak, I missed that discussion.
Would you support a PR that changes FSAC to do the cache initialization lazily?

Krzysztof-Cieslak:
> Would you support a PR that changes FSAC to do the cache initialization lazily?

👍

olivercoad:
@Krzysztof-Cieslak can you expand on why you closed this PR?

> As described before (#1041 (comment)) this is by design behaviour.

Yes - the whole point of this PR was to change that behaviour. Wouldn't the other activation events be sufficient without *.sln?

travv0:
> 
> 
> As described before ([#1041 (comment)](https://github.com/ionide/ionide-vscode-fsharp/issues/1041#issuecomment-576668861)) this is by design behaviour.

There's no mention in that issue of *why* that's the intended behavior.  I can't think of any reason why that would be the case.

smcl:
Is there any way we could adjust this? It seems to me that there are a sufficient collection of activation events to cover the common F# use cases without requiring activation on seeing `.sln` files

PR #1403, opened 2020-07-21
add a publish task that triggers on tags only

baronfel:
this task publishes the extension from the release directory on windows tagged builds only.

If you've got a secret called VSCE_PAT in azure devops, then I think this should just work!


Krzysztof-Cieslak:
Can we move all that stuff to the GH Actions?

Krzysztof-Cieslak:
Also, it shouldn't run yarn manually - we have FAKE target for release.

Krzysztof-Cieslak:
My local release cmd looks like this - it should be same flow for release here. 
```
@echo off
cls

SET nuget-key=xxx
SET github-user=Krzysztof-Cieslak
SET github-pw=xxx
SET vsce-token=xxx

dotnet tool restore
dotnet fake build --target Release
```

baronfel:
Cool, lemme replicate that here then. I thought about GH actions, but they don't easily support having multiple .net sdks/runtimes and that's a problem we have pretty often in this repo, so I thought it better to keep things in azdo.

If we did move to GH actions we could also split up MR pipelines from tag pipelines in a much nicer way than the tag-reference I have here.

baronfel:
alright, the calls are the same now. you'll still need to set up azdo secrets with those same names, though.

baronfel:
Ok, I've got a parallel github actions-based pipeline now, that seems to be doing the things. Want me to drop the azdo one?

PR #1405, opened 2020-08-10
update FSAC to get new preview features and fixes

baronfel:


Zaid-Ajaj:
Hi @baronfel, FSAC was updated with latest FCS 37 but the FSharp.Analyzers.SDK hasn't yet, does it say backward-compatible with FSAC even after updating the compiler services?

baronfel:
to _some_ degree, but not all. I was going to come back and start those upgrades now 👍 depending on what your analyzer accesses it can be fine, but if it hits a new AST node (ie interpolated string) matchexceptions may get thrown, type mismatch exceptions, etc.

But in the general case, the analyzer tests in FSAC passes so there's not a showstopper critical breakage as of yet.

Zaid-Ajaj:
Thanks for the quick answer Chet :heart: Alright, then fingers crossed that it doesn't break 🤞 

However, can I assume it *will* break if we update the analyzer sdk, say to 0.6.0 and my analyzer depends on sdk 0.5.0?

baronfel:
that's likely, yeah. I was going to try to hit both the sdk and your project today as a matter of fact :)

Zaid-Ajaj:
That would be awesome Chet :heart: 😍 thanks a lot!

PR #1415, opened 2020-09-13
Upgrade Fable

### LINKED ISSUE #1036: Upgrade to Fable 2
@Krzysztof-Cieslak Would you be ok, if I upgrade the repo to Fable 2.1 ?

In theory it will:
- Decrease the bundle size
- Improve the performance
- Be up to date again with latest Fable version
- Easier to start as `dotnet fable` is now replace with `webpack`

Booksbaum:
Fixes #1036 (though a couple of versions higher)

There are a couple of changes in the current fable version compared to the one used in current ionide -- and some are quite subtle. For example passed functions that previously were called now don't (ionide/ionide-vscode-helpers@ada99c8728c372994c0a8a00136bc0b23adbed28) or stricter handling for different types in F# (in code) and JS (during runtime) (92ecbd01340ae5f05a7b630330286bcf4c9b0ae9).

I'm not sure I caught all those strange cases. The ones I found had strong impact -- like no files in Solution Explorer. But there might be other cases not so visible. And I didn't test it with a bigger project and over a longer period of time.  
-> others should probably test it a bit more with their workflow

Requires ~~ionide/ionide-vscode-helpers#33~~ and ~~MangelMaxime/html-to-elmish#22~~ ~~(-> reason this PR is marked as *WIP*)~~  
For now the files for both are included from the PRs branches. So this should build just fine for testing.
I will change it back once the PRs are merged (and updated nuget package is released in case of `html-to-elmish` / `Fable.HtmlConverter`).


Krzysztof-Cieslak:
Can you target `5.0` branch with this PR.

Booksbaum:
`ionide-vscode-helpers` are from the ionide repo again. And `Fable.HtmlConverter` is updated and included as nuget package.  
-> good to go

cartermp:
Some of the refactoring done (e.g., aliasing `Func<obj, obj>`) feels a little churn-y but otherwise these changes appear reasonable

Krzysztof-Cieslak:
Well, this is awesome, thanks a lot ❤️

Issue #1176, opened 2019-07-11
[Enhancement] Info Panel: Only auto-popup when an F#-type file is opened.


baronfel:
**Is your feature request related to a problem? Please describe.**
When I open my work repository and I was last working on yaml/bash/any non-F# file, the Info Panel pops up.  When I'm working with F# files I want this behavior, but for non-F# coding tasks I don't want it right away.

**Describe the solution you'd like**
I'd like if the Info Panel wasn't popped-up automatically until the first F# file was opened.  This would keep the Info Panel relevant while minimizing the number of times it's present and not immediately useful, I think.  Imagine the case where you open a yaml file first: the info panel is present and completely blank, obscuring parts of your screen. Now imagine a world where you work in yaml unobstructed until you go to some F# code and BAM your beloved Info Panel is right there with you, helping you along your coding journey.

**Describe alternatives you've considered**
Closing the Info Panel all the time. This hurts because it's like I'm saying goodbye to a close friend.

Krzysztof-Cieslak:
Makes sense, should be fairly easy to do - in the place where we make InfoPanel visible during plugin startup we can check current list of opened documents and check if any is F# file.

teo-tsirpanis:
Another option is to never auto-popup, or even better, to configure its behavior through the settings.

MangelMaxime:
> Another option is to never auto-popup, or even better, to configure its behavior through the settings.

You can already configure it's behaviour via the settings:

<img width="588" alt="Capture d’écran 2019-07-15 à 12 18 34" src="https://user-images.githubusercontent.com/4760796/61209683-b3cb7100-a6fa-11e9-8f88-6da143acd26e.png">

teo-tsirpanis:
That’s exactly how my setting look like. Doesn’t the second option being disabled imply that the F# window shouldn’t popup on start?

Από: Maxime Mangel
Αποστολή: Δευτέρα, 15 Ιουλίου 2019 1:19 μμ
Προς: ionide/ionide-vscode-fsharp
Κοιν.: Theodore Tsirpanis; Comment
Θέμα: Re: [ionide/ionide-vscode-fsharp] [Enhancement] Info Panel: Onlyauto-popup when an F#-type file is opened. (#1176)

Another option is to never auto-popup, or even better, to configure its behavior through the settings.
You can already configure it's behaviour via the settings:

MangelMaxime:
[...]

PR #1438, opened 2020-11-23
Only show info panel for fsharp files

wilsoncg:
As discussed in #1176 

I noticed 2 issues when opening C# projects with vscode after installing ionide extension:
1. With `"FSharp.infoPanelShowOnStartup": true`, InfoPanel launches for a C# project & displays a blank panel (because of .sln activation). Closing the panel, then closing & re-opening C# project in vscode does not resolve the issue. A blank info panel is displayed.
2. With `"FSharp.showExplorerOnStartup": true`, when launching a C# project the F# solution explorer triggers it's startup, grabs focus but then reports a failure in attempting to load csproj files.

A more ideal behaviour would be to try and have a minimal ionide fsharp launch when opening C# projects.

This fix resolves issue 1), by only showing the InfoPanel if there is an fsharp compatible file open in a window. I imagine issue 2) is by design, and can easily be resolved by setting `"FSharp.showExplorerOnStartup": false`.

Behaviour before code change:
ionide-infopanel-displaying-for-csharp

Behaviour after code change:
ionide-infopanel-only-for-fsharp

Thanks to @baronfel for excellent instructions in #1176 👍

### DIFF
--- a/src/Components/InfoPanel.fs
+++ b/src/Components/InfoPanel.fs
@@ -14,6 +14,15 @@ module node = Fable.Import.Node.Exports
 
 module InfoPanel =
 
+    let private isFsharpTextEditor (textEditor : TextEditor) =
+            if JS.isDefined textEditor && JS.isDefined textEditor.document then
+                let doc = textEditor.document
+                match doc with
+                | Document.FSharp -> true
+                | _ -> false
+            else
+                false
+
     module Panel =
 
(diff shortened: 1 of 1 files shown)


baronfel:
Lovely, thanks for this!  I just pushed a preview ahead of the US holiday coming up, but if I get a moment I'll see about pushing out another.

PR #1448, opened 2020-12-09
delegate to lsp for documentation

baronfel:
This is a prototype for using `workspace/executeCommand` to invoke commands on the backend that can cause client changes.

Instead of creating the text edits in ionide proper, we can have the LSP implementation in FSAC send text edit commands back to the client.

Marking as draft while I continue discussion in FSAC.


baronfel:
Closing as there's something about this mechanism I'm not understanding at all.

PR #1451, opened 2020-12-11
better check for if a project is an executable

baronfel:
You can't launch or debug net5.0 projects right now because the check we were doing to determine if a project was an app didn't take net5.0 into account.

I changed the logic to just use the structures that we get from FSAC instead of doing manual project file parsing.  Now the buttons work!

### DIFF
--- a/src/Core/Project.fs
+++ b/src/Core/Project.fs
@@ -53,21 +53,10 @@ module Project =
 
     let deepLevel = "FSharp.workspaceModePeekDeepLevel" |> Configuration.get 2 |> max 0
 
-    let isANetCoreAppProject (project : Project) =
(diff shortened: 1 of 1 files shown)


src/Core/DTO.fs
```diff
@@ -237,7 +237,7 @@ module DTO =
           References : ProjectReferencePath array
           Logs : Map<string, string>
           OutputType : string
-          Info : ProjectResponseInfoDotnetSdk
+          Info : {| Data: ProjectResponseInfoDotnetSdk |}
```
baronfel, line 240: not sure if we intended for the data structure to change here, but this was causing several problems in various places when trying to use project data.

Issue #1416, opened 2020-09-15
Add type hints for last line of a pipeline


forki:
the new type hints in pipelines are an excellent time saver. I really love them. 

One nice addition would be to add type hints for the last line of a pipe as well. 
So we can see the type without hovering over the let binding. 
This is especially useful for local pipelines (inside a function) since Ionide doesn't show type hints for those bindings.

Krzysztof-Cieslak:
Fixed by https://github.com/ionide/ionide-vscode-fsharp/pull/1453

PR #1453, opened 2020-12-19
Expand pipeline hints to final line and improve position

tboby:
This is the ionide side of implementing this improvement: ionide/ionide-vscode-fsharp#1416
See description in https://github.com/fsharp/FsAutoComplete/pull/683

This just amends the processing of the return to create two type hints where appropriate.

I haven't updated the lock file, I'm not sure how to co-ordinate syncing this with FSAC!

The two effects are moving pipeline hints to the correct line when there are comments, and adding the type hint on the final line.
image

### DIFF
--- a/src/Components/PipelineHints.fs
+++ b/src/Components/PipelineHints.fs
@@ -118,19 +118,31 @@ type State =
 
 module DecorationUpdate =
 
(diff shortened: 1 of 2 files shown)


baronfel:
Generally as soon as an FSAC MR that needs ionide-sode changes as merged you'd paket update on the ionide side, which will update the FSAC reference. We can worry about that after the MR though.

PR #1465, opened 2020-12-29
Piepline to Pipeline

voronoipotato:
It's not that important but it does make things a little more searchable


baronfel:
Thanks!

cartermp:
I always searched with PieplineHint so this is a breaking change

PR #1468, opened 2020-12-30
Fix #1460: Update to Fable 3

### LINKED ISSUE #1460: Update to Fable 3.0

Booksbaum:
fixes #1460

Requires ionide/ionide-vscode-helpers#35

About Experimental Build:  
It seems like experimental builds aren't done via the build script, but by hand with probably a command like `webpack --mode=production --env.ionideExperimental` (which passes `IONIDE_EXPERIMENTAL` to Fable).  
With Fable 3, compilation to JS and bundling are now split in two between Fable and webpack. Which makes the command more complicated.  
I added a Build Target to for that: `BuildExp` -- existed already before, but was empty. Now it compiles with Fable with `IONIDE_EXPERIMENTAL` defined, and bundles with webpack and `--env.ionideExperimental` into `release-exp`. As far as I can tell that's how an experimental build was created before.  
If experimental build should be kept out of the build script (or requires more options): The full command is now:   
`dotnet fable ./src/Ionide.FSharp.fsproj --outDir ./out --define IONIDE_EXPERIMENTAL --run webpack --mode=production --env.ionideExperimental`

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -9,9 +9,15 @@
       ]
     },
     "paket": {
-      "version": "5.249.2",
+      "version": "5.257.0",
       "commands": [
         "paket"
+      ]
+    },
+    "fable": {
+      "version": "3.0.5",
+      "commands": [
+        "fable"
       ]
     }
   }
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,4 @@ types.txt
 .ionide
 .fake
 fsacpath.txt
+/out/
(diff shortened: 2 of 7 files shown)


Krzysztof-Cieslak:
I think Experimental build is not needed - it was used back before 4.0 release, and we're not using it any more.

I've merged the helpers PR so you can bump dependencies here.

Krzysztof-Cieslak:
Awesome, thanks a lot for this ❤️

forki:
Wow. @alfonsogarciacaro did you see this?

alfonsogarciacaro:
Sorry, didn't I tell you that all CLI options are changing in Fable 3.1? ...Just kidding 😜 Great work @Booksbaum!

PR #1471, opened 2021-01-01
Fail promise if msbuild has exit code > 0

tboby:
If "F#: Debug Default project" fails to build, it currently continues and runs the previous build which can be misleading.
This is a crude approach that just stops and tells the user it hasn't worked.

To handle this properly (e.g. focus the error output), `invokeMSBuild` should handle the failure, but that requires pushing the handling done by some other consuming functions up/down. Should I have a go at doing this?


Krzysztof-Cieslak:
Thanks!

Issue #1456, opened 2020-12-21
ionide 5.0 - error by moving F# files in F# view


vilinski:
**Describe the bug**

Can't reorder F# source files in the F# project view using arrows

```
Error running command fsharp.explorer.moveUp: System.NullReferenceException: Object reference not set to an instance of an object.
   at FsAutoComplete.FsProjEditor.moveFileUp(String fsprojPath, String file) in D:\Programowanie\ionide\paket-files\github.com\fsharp\FsAutoComplete\src\FsAutoComplete.Core\FsprojEdit.fs:line 11
   at <StartupCode$FsAutoComplete-Core>.$Commands.FsProjMoveFileUp@363.Invoke(Unit unitVar) in D:\Programowanie\ionide\paket-files\github.com\fsharp\FsAutoComplete\src\FsAutoComplete.Core\Commands.fs:line 363
   at Microsoft.FSharp.Control.AsyncPrimitives.CallThenInvoke[T,TResult](AsyncActivation`1 ctxt, TResult result1, FSharpFunc`2 part2) in F:\workspace\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 386
   at Microsoft.FSharp.Control.Trampoline.Execute(FSharpFunc`2 firstAction) in F:\workspace\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 104. This is likely caused by the extension that contributes fsharp.explorer.moveUp.
```

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to F# project view with opened F# project with more than one source file
2. Click on reorder arrow - up or down
3. See error

**Expected behaviour**
The F# source file is moved withing project files order.

**Environment (please complete the following information):**
 - OS: MacOS Big SUR
 - Ionide version: 5.0
 - VSCode version: 1.52.1
 - dotnet SDK version: 5.0.100
 - mono / .Net Framework version: none

**Additional context**
Dunno whether important - upgraded from 5.0 preview 03, cleared project cache, reloaded vscode window. Can't move files

Krzysztof-Cieslak:
Are the files you try to move in a folder by any chance?

vilinski:
Yes! Just tried again - it works if filesa are flat in a project root, but not in subfolders.
An also usual constellation with Shared folder where files are shared between Client and Server projects throws NPE

rodsfreitas:
I'm having a possible related issue, when trying to add a new file from F# Explorer. 

[...]

PR #1472, opened 2021-01-03
Solution explorer: Compile order and better support for directories

tboby:
This PR makes two changes to the Solution Explorer TreeView:
1. Compile order layout: The solution explorer tree now show shows all files in the same order they are listed in the fsproj by creating duplicate folder nodes. To do this it assumes that the list of Items provided by FSAC is in compile order (which seems to be the case).
Before: 
image
After;
image
2. Improved file actions in subdirectories (https://github.com/ionide/ionide-vscode-fsharp/issues/1456): In combination with changes in FSAC, this improves actions on non-root directory files by storing the "virtual path" and passing this to the FSAC commands rather than just the file name.
a. Add file above/below now prompt for a file name relative to the selected file
b. Add file is still relative to project root
c. Move up/down still don't behave consistently with mixed path separators though


tboby:
@Krzysztof-Cieslak Any feedback?

Krzysztof-Cieslak:
Thanks!

PR #1475, opened 2021-01-05
use the new ranges from the updated FSAC endpoint

baronfel:
This depends on https://github.com/fsharp/FsAutoComplete/pull/693, but updates Ionide to use the new API types exposed as part of that work.  Casing of some things has changed, as the payload we get back from the request is no longer a bare string, instead it's already been hydrated as a JS object due to the LSP back-and-forth.


baronfel:
alright, this is ready as well

cartermp (commented):
yeet it in

cartermp (approved):

PR #1491, opened 2021-02-01
update vscode deps so that semantic tokens are supported

baronfel:
This is the companion MR to https://github.com/fsharp/FsAutoComplete/pull/711.

In order to support semantic tokens at the LSP layer, we have to update our vscode-languageserver dependency to one that knows about Semantic Tokens.

We also have to remove our shim for semantic tokens to allow the LSP layer to take it over.

I also added a new flag to the debugger attachment code for FSAC, because `--attachDebugger` didn't seem to work on MacOS.


baronfel:
This should not be merged until the associated FSAC PR is merged, and then this should include the FSAC update as part of it.

PR #1492, opened 2021-02-06
Open visx

nojaf:
You should totally use #1484 instead of this half baked PR.
Other than that, I didn't see that @muhammadsammy already had a PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,3 +49,11 @@ jobs:
         vsce-token: ${{ secrets.VSCE_TOKEN }}
         user-email: ${{ github.event.inputs.email }}
       run: dotnet fake build -t Release
+    - name: Build OpenVsix package
+      run: dotnet fake build -t BuildPackgeOpenVsix
+    - name: Publish to Open VSX Registry
+      uses: HaaLeo/publish-vscode-extension@v0
+      id: publishToOpenVSX
+      with:
+        packagePath: "./temp"
+        pat: ${{ secrets.OPEN_VSX_TOKEN }}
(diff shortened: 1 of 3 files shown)


muhammadsammy:
@nojaf My PR isn't complete either, I was experimenting and wasn't sure it would be useful because I do not know F#.
So It would be better if someone who actually knows F# took this over. Thank you!

baronfel:
This one looks pretty close, I think the thing that's missing is we'd need to ensure that the open-vsix tool is installed on the build agents.

nojaf:
I'm actually assuming that the GitHub action is all you need to publish.
There is no difference in building the package as far as I'm aware.
I was able to install the built package in gitpod and that worked.

baronfel:
oh of course, I completely missed that you were using an action specifically for that. So my understanding would be that we'd merge this, @Krzysztof-Cieslak could register on open-vsix, and then a test release could be run, then?

nojaf:
Yes, however, OpenVsix requires that you "claim" a namespace before you can push a package.
See https://github.com/eclipse/openvsx/wiki/Namespace-Access#how-to-claim-a-namespace

baronfel:
@Krzysztof-Cieslak has already registered and claimed on ovsx: https://github.com/EclipseFdn/open-vsx.org/issues/134, and we already have a secret configured here. So we should be able to merge this now.

baronfel:
Had to fix a slight issue in the fake logic (my bad for not catching that as part of this MR, but we're published on Open-vsx now! https://open-vsx.org/extension/Ionide/Ionide-fsharp

Issue #1490, opened 2021-01-31
Add Current File To Project


mohaalak:
I'm new to F# I was watching some videos on youtube, the developer runs a command called 'Add current File To Project', I did not see this command anywhere, I was searching and found out there was a tool called Forge that is now deprecated that it was this tool job to add the file to project and move up and down the files.

I don't know if this command changed to something else or not

also, I'm reading the project right now if you think that this is a beginner task I can do it and send a pull request.

v-zubritsky:
@mohaalak This doesn't answer your question, just to add some information. 
Forge is indeed [deprecated](https://github.com/ionide/Forge/issues/135) just like you said, while file related operations are provided by [FsAutoComplete](https://github.com/fsharp/FsAutoComplete/blob/master/src/FsAutoComplete.Core/FsprojEdit.fs).

I might well be mistaken.

baronfel:
You've got it correct @v-zubritsky. In addition, recently some contributors have massively reworked the file support in Ionide, so I'd check and see if your request is now part of the product.

mohaalak:
there is also a bug just create a f# file but don't add it into fsproj this will show up in the status bar but when you click on it there is a message that says`fsharp.AddFileToProject not found`

<img width="794" alt="Screen Shot 2021-02-07 at 10 30 55 PM" src="https://user-images.githubusercontent.com/891975/107156721-4692c700-6995-11eb-8fb7-50937f479560.png">

I'm working on a pull request right now I found the solution using this [function](https://github.com/ionide/ionide-vscode-fsharp/blob/e68e92da62a1437b7c2fdab5f542ddbc8b210758/src/Core/FsProjEdit.fs#L33)

Issue #1479, opened 2021-01-13
ionide does not load projects


atheck:
### Describe the bug**

I'm using ionide in vs code. After opening vs code ionide does not load any projects. All projects are "not loaded yet". When I open a fs file, in the status bar it says "Not in a F# project" and the error is: "command 'fsharp.AddFileToProject' not found".

Installed dotnet SDKs:
* 2.1.812 [/usr/share/dotnet/sdk]
* 3.1.405 [/usr/share/dotnet/sdk]
* 5.0.102 [/usr/share/dotnet/sdk]

Installed dotnet runtimes:
* Microsoft.AspNetCore.All 2.1.24 [/usr/share/dotnet/shared/Microsoft.AspNetCore.All]
* Microsoft.AspNetCore.App 2.1.24 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
* Microsoft.AspNetCore.App 3.1.11 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
* Microsoft.AspNetCore.App 5.0.2 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
* Microsoft.NETCore.App 2.1.24 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
* Microsoft.NETCore.App 3.1.11 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
* Microsoft.NETCore.App 5.0.2 [/usr/share/dotnet/shared/Microsoft.NETCore.App]

Also I have installed latest mono.

Am I missing something?

### Expected behaviour
All projects should load.

## Machine info
 - OS: Ubuntu 20.04
 - .NET SDK version: see above
 - Ionide version: 5.1.0

atheck:
Hi, it is working again. I don't know why. Maybe some update.

Issue can be closed.

rytido:
I am having this same issue on Windows with the same SDKs installed. @atheck no idea why it started working again?

atheck:
@rytido no sorry, I don't know why it's working again. I did nothing but ignoring it. Maybe we should reopen this issue?

rytido:
I think so. I uninstalled all the SDKs and Ionide and reinstalled, and I still get the same issue. Thank you.

SluttyPutty:
I unsubbed this a while ago. why am i still receivinf these.

On Jan 22, 2021, at 11:06, rytido <notifications@github.com> wrote:

﻿
[...]

PR #1497, opened 2021-02-10
Add Command: Add current file to project

mohaalak:
This pull request will fix the issue that is mentioned on #1490 also on #1479 (but not entirely).
this issue was raised after the Forge tool was removed and a command named AddFileToProject dependent on it was removed.


mohaalak:
with this pull request when the user is on a fsharp file, he can use 'Add Current File To Project' command, it will show a list of projects that he can choose from and it will be added to the project.

Krzysztof-Cieslak:
Thanks!

PR #1510, opened 2021-03-03
bump FSAC to pick up FCS 39 and cracking improvements

baronfel:


baronfel:
This is working great for me on a locally-built vsix, just doing a bit more testing before merging and pushing a new version.

Zaid-Ajaj:
@baronfel Would be happy to test it on my repositories if you can upload the vsix here :smile:

baronfel:
Here you go! You'll have to remove the `.zip` extension to get it to work (thanks Github uploads!)

[Ionide-fsharp-5.4.0.vsix.zip](https://github.com/ionide/ionide-vscode-fsharp/files/6079081/Ionide-fsharp-5.4.0.vsix.zip)

Zaid-Ajaj:
Awesome! I will let you know how it goes :smile:

cartermp (approved):
Works on my machine.

Zaid-Ajaj:
Works on my machine too but the perf for a larger repo remains 😢  It felt like it was loading faster for the a small project of 1k lines of code and was very responsive so that's great, but when I open Fable.Remoting the perf feels similar to what it was before. Loading the projects does work and I get all the codelens and tooltip goodies but to the point where the editor itself loses responsiveness and I can't even scroll (due to memory pressure? I don't know) 

2021-03-03 22_12_41-Task Manager

Although this is somewhat workable for me and I can manage my way around, these projects that I am using are nothing fancy and are relatively small. I can only imagine how it will impact bigger repos. That said, my laptop isn't that much a powerhouse but it isn't too shabby either (I think) at least for dev. 

2021-03-03 22_22_56-Control Panel_System and Security_System

baronfel:
Ok, I'm going to go ahead with this because it un-breaks peoples' situations, but we should follow up with you on the perf side of things (also Jimmy Byrd had some ideas around some perf stuff in FSAC proper that we plan on running down).

Zaid-Ajaj:
Like I mentioned, it could be just me where my laptop can't handle what fsac is doing. I am curious as to how the experience looks like for you when open [Fable.Remoting](https://github.com/Zaid-Ajaj/Fable.Remoting) locally. Can you maybe try it out and see how Ionide performs against it? for example whether editor freezes and memory usage spikes are noticeable

baronfel:
I can't reproduce the hanging/memory usage. On my laptop, the two largest dotnet processes are 1.7GB and 1.4 GB, and the editor experience is quite smooth. Weirdly so, honestly.

cartermp:
We'll need to set up a proper repro @Zaid-Ajaj since I also don't see that level of memory or CPU usage after building the repo --> `code .` --> opening some files.

Zaid-Ajaj:
@cartermp @baronfel thanks a lot for checking! I tried this again, this time I deleted `.ionide` and ran a test target before opening the Fable.Remoting repository and it actually loaded A LOT faster, the codelenses and tooltips were really responsive too 😄 I am not sure what deleting `.ionide` does but it seemed to have helped. Repeating the same process for the Feliz yielded similar results: the project is responsive and the editor is not hanging anymore 🎉

PR #1511, opened 2021-03-04
Fix auto comment indentation

MangelMaxime:
When pressing enter on a `///` comment line the indentation is incorrect

**Before**
vscode_comment_indent_problem

**After**
vscode_comment_indent_problem_after


baronfel:
Oh this is lovely, thank you

PR #1513, opened 2021-03-09
Add option to crack using msbuild project graph

TheAngryByrd:
This PR is a follow through from ionide/proj-info#102 and https://github.com/fsharp/FsAutoComplete/pull/736 to allow MSBuild ProjectGraph WorkspaceLoader. This pass a flag on the CLI to FSAC to turn on the ability to use this new loader while defaulting to the current one.


baronfel:
Thanks! looking forward to trying this out

PR #1551, opened 2021-07-25
reenable sourcemaps and script-file xml generation, and add in escape hatch for .net 6 support

baronfel:
Couple of fixes I've accumulated:

* reenable sourcemaps for local debugging
* remove an artificial constraint around xml doc comment generation that prevented it from working on script files
* Add in `FSharp.fsac.dotnetArgs` as a way for users to specify addition args to the `dotnet` cli itself. the primary intended purpose will be to set `--fx-version` to whatever their latest installed runtime is until we provide a .net 6-first compatible build.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -15,7 +15,7 @@
       ]
     },
     "fable": {
-      "version": "3.0.5",
+      "version": "3.2.9",
       "commands": [
         "fable"
       ]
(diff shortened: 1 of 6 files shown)


baronfel:
For the release notes, we'll want to give an example of getting the .net 6 runtime and applying it using these args

.config/dotnet-tools.json
```diff
@@ -1,24 +1,24 @@
...
+      ]
+    },
+    "paket": {
+      "version": "5.257.0",
+      "commands": [
+        "paket"
+      ]
+    },
+    "fable": {
+      "version": "3.2.9",
```
baronfel, line 18: fable 3.2 brought sourcemaps back

PR #1553, opened 2021-07-28
massive refactor/update for new vscode types

baronfel:
This is a WIP to see what the impact of the types from https://github.com/ionide/ionide-vscode-helpers/pull/39 is on this codebase.

So far, a lot of cheese-moving but nothing incredibly scary.

The key points of concern I had are:

* if the `objfy` functions are working correctly
* if the Thenable/Promise interop (see Utils.fs) is working correctly

In local testing things seem to be working rather well, but since the surface area of this is pretty large there are certain to be gaps.


src/Components/CodeLensHelpers.fs
```diff
@@ -2,26 +2,27 @@ namespace Ionide.VSCode.FSharp
...
             )
             |> ResizeArray
-        commands.executeCommand("editor.action.showReferences", uri, pos, locs )
+        commands.executeCommand("editor.action.showReferences", ResizeArray [Some (box uri); Some (box pos); Some (box locs);])
 
     let activate (context : ExtensionContext) =
 
-        commands.registerCommand("fsharp.showReferences", showReferences |> objfy4 ) |> context.subscriptions.Add
\ No newline at end of file
+        commands.registerCommand("fsharp.showReferences", showReferences |> objfy4 ) |> box |> unbox context.subscriptions.Add
```
baronfel, line 28: you'll see this pattern _a lot_ because the types from `registerCommand` (IDisposable) don't line up with the types of `context.subscriptions.Add` (aka `ResizeArray<ExtensionContextSubscriptions>.Add`, aka ExtensionContextSubscriptions -> unit`).

Disposable and ExtensionContextSubscriptions have the same shapes:

```fsharp
    type [<AllowNullLiteral>] ExtensionContextSubscriptions =
        abstract dispose: unit -> obj option
```

```fsharp
    /// Represents a type which can release resources, such
    /// as event listening or a timer.
    type [<AllowNullLiteral>] Disposable =
        /// Dispose this object.
        abstract dispose: unit -> obj option
```

but we need some way to coerce between them (and other semantically-similar types) in a safe way.

src/Core/Utils.fs
```diff
@@ -1,7 +1,16 @@
...
-open Fable.Import.vscode
+open Fable.Import.VSCode
+open Fable.Import.VSCode.Vscode
+open Fable.Core
+
+[<AutoOpen>]
+module PromiseBuilder =
+    type Foo = JS.PromiseConstructor
+    type Promise.PromiseBuilder with
+        member x.Source(p: Thenable<'t>): JS.Promise<'t> = box p |> unbox
```
baronfel, line 12: adding these lets us await Thenables.  we also have to add the self-referential Source overload or else overload resolution fails.  this 'default' Source should be contributed upstream.

src/Core/Utils.fs
```diff
@@ -355,18 +370,22 @@ module VSCodeExtension =
...
         sprintf "workbench.view.extension.%s" extensionName
 
 
 [<AutoOpen>]
 module Objectify =
-    let inline objfy2 (f: 'a -> 'b): obj -> obj = unbox f
-    let inline objfy3 (f: 'a -> 'b -> 'c): obj -> obj -> obj = unbox f
-    let inline objfy4 (f: 'a -> 'b -> 'c -> 'd): obj -> obj -> obj -> obj = unbox f
-    let inline objfy5 (f: 'a -> 'b -> 'c -> 'd -> 'e): obj -> obj -> obj -> obj -> obj = unbox f
+    let inline objfy2 (f: 'a -> 'b): ResizeArray<obj option> -> obj option = unbox f
```
baronfel, line 388: all the commands-taking functions now want a resizearray of parameters

src/Components/CodeLensHelpers.fs
baronfel: I added a small SRTP member to the ExtensionContext type that cleared up a lot of this noise

baronfel:
using the new terminal profile provider mechanism we have a nice menu entry now: 
image

baronfel:
merging now that the typings are stabilized and we want to do more features on top of them

PR #1556, opened 2021-07-28
clear out old terminal state when a new one is created

baronfel:
tweaks terminal invalidation logic to clear out _as_ a new terminal is being created. this results in pulling some logic out of start/close and into the new 'terminal started' event handler.

### DIFF
--- a/src/Components/Fsi.fs
+++ b/src/Components/Fsi.fs
@@ -425,15 +425,16 @@ module Fsi =
             |> Promise.suppress
             |> ignore)
 
-    let private handleCloseTerminal (terminal : Terminal) =
-        ()
 
     let private clearOldTerminalState () =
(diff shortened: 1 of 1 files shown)


src/Components/Fsi.fs
```diff
@@ -425,18 +426,23 @@ module Fsi =
             |> ignore)
 
     let private handleCloseTerminal (terminal : Terminal) =
-        fsiOutputPID
```
Krzysztof-Cieslak, line 428: `handleCloseTerminal` needs to reset the state. Otherwise, we are broken for following scenario:
1. Close terminal
2. Run `FSI: Send Selection`

baronfel: theres a bit of a race condition with closing it out during startup then, I'd imagine

PR #1559, opened 2021-08-04
set rollforward args and env vars if .net 6.0 is detected

baronfel:
This PR detects the dotnet SDK version in the rootPath and provides arguments and environment variables to add to the FSAC invocation if the detected SDK is 6.0+ or a prerelease.

This should eliminate most users' need to set the fsac.dotnetArgs parameter entirely and provide a seamless experience while we don't ship a multitargeted FSAC.

### DIFF
--- a/src/Core/LanguageService.fs
+++ b/src/Core/LanguageService.fs
@@ -8,6 +8,7 @@ open Fable.Import.VSCode
 open Fable.Import.VSCode.Vscode
 open global.Node
 open Ionide.VSCode.Helpers
+open Semver
 
 open DTO
 open LanguageServer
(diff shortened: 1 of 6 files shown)


src/Core/LanguageService.fs
```diff
@@ -72,6 +73,23 @@ module LanguageService =
...
+    let runtimeVersion() = promise {
+        let! dotnet = Environment.dotnet
+        match dotnet with
+        | None -> return Core.Error "No dotnet binary found"
+        | Some dotnet ->
+            let! (error, stdout, stderr) = Process.exec dotnet "" "--version"
+            match error with
+            | Some e -> return Core.Error $"error while invoking: {e.message}"
+            | None ->
+                let stdoutTrimmed = stdout.TrimEnd()
```
baronfel, line 86: need to trim because it contains a newline at the end

src/Core/LanguageService.fs
```diff
@@ -650,13 +668,73 @@ Consider:
...
+
+            printfn $"FSAC (NETCORE): '%s{fsautocompletePath}'"
+
+            let exeOpts = createEmpty<ExecutableOptions>
+            let exeEnv =
+                match fsacEnvVars with
+                | [] -> None
+                | fsacEnvVars ->
+                    // only need to set the process env if FSAC needs rollfoward env vars.
+                    let keys = Node.Util.Object.keys process.env
```
baronfel, line 729: we seemingly have to copy over the current process environment, even though my reading of the source code suggests otherwise.  if I didn't do this then cracking failed every time.

src/Core/LanguageService.fs
```diff
@@ -650,13 +668,73 @@ Consider:
...
+                    let baseEnv = keys |> Seq.toList |> List.map (fun k -> k, process.env?(k))
```
baronfel, line 730: super subtle fable-ism here: `process.env?k` returns `undefined` (it's like calling `process.env.k`, but `process.env?(k)` succeeds (it's like calling `process.env[k]`)

Issue #1560, opened 2021-08-05
Version 5.7.0: Running the contributed command: 'fsi.SendSelection' failed.


dgrimmin:
### Describe the bug**
With the release of version 5.7.0 sending script to fsi stopped working.

### Steps to reproduce
Open script select text and alt+enter

### Expected behaviour
terminal opens and script gets executed

## Machine info
 - OS: Windows
 - .NET SDK version: 5
 - Ionide version: 5.7.0

### Workaround
Revert to 5.6.0

baronfel:
Thanks for reporting! I introduced a regression in this release, I'll try to take a hotfix today to resolve it. In the meantime you can manually use the previous version by choosing it from the extensions page in vscode.

andagr:
Log (copied from the [discussion about this in slack](https://fsharp.slack.com/archives/C04BJKUP2/p1628143865131100)):

```
mainThreadExtensionService.ts:63 [[object Object]]Terminal has already been disposed
$onExtensionRuntimeError @ mainThreadExtensionService.ts:63
_doInvokeHandler @ rpcProtocol.ts:418
_invokeHandler @ rpcProtocol.ts:403
_receiveRequest @ rpcProtocol.ts:319
_receiveOneMessage @ rpcProtocol.ts:246
(anonymous) @ rpcProtocol.ts:111
fire @ event.ts:583
fire @ ipc.net.ts:513
_receiveMessage @ ipc.net.ts:866
(anonymous) @ ipc.net.ts:705
fire @ event.ts:583
acceptChunk @ ipc.net.ts:286
(anonymous) @ ipc.net.ts:247
T @ ipc.net.ts:50
emit @ events.js:315
addChunk @ internal/streams/readable.js:309
readableAddChunk @ internal/streams/readable.js:284
Readable.push @ internal/streams/readable.js:223
onStreamRead @ internal/stream_base_commons.js:188
mainThreadExtensionService.ts:64 Error: Terminal has already been disposed
	at S._checkDisposed (c:\Users\user\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:85:32647)
	at Object.sendText (c:\Users\user\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:85:31253)
[...]

PR #1561, opened 2021-08-05
remove duplicate state initialization in FSI start loop

baronfel:
When rebasing the standalone terminal fixes back into the new vscode typings PR I messed up the initialization routine in FSI.fs

The `start` method now starts the terminal with a given profile, and the `handleOpenTerminal` handler is used to ensure that any started terminal is cleanly initialized and all mutables are updated accordingly.  This is necessary because we use the same shared terminal profile now when users manually start FSI via the terminal dropdowns as we do when they invoke FSI via any of the FSI commands.

This was the case in master before the typings MR as well.

### DIFF
--- a/src/Components/Fsi.fs
+++ b/src/Components/Fsi.fs
@@ -320,7 +320,6 @@ module Fsi =
         )
 
     let private start () =
-        fsiOutput |> Option.iter (fun n -> n.dispose())
         promise {
             let ctok = vscode.CancellationTokenSource.Create().token
             let! profile =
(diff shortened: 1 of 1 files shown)


src/Components/Fsi.fs
```diff
@@ -320,7 +320,6 @@ module Fsi =
         )
 
     let private start () =
-        fsiOutput |> Option.iter (fun n -> n.dispose())
```
baronfel, line 323: all of this state management (cleanup of old terminal, tracking of new values) is handled in the `handleOpenTerminal` event now, which is fired after this terminal is created.

baronfel:
Closes #1560

PR #1564, opened 2021-08-07
track terminals based on their file and cwd context to remove the need to manage mutables

### LINKED ISSUE #1562: Invalid __SOURCE_DIRECTORY__ on first "FSI: Send Selection"
### Describe the bug**

When sending selection to FSI the first time (Ionide opens fsi terminal) the __SOURCE_DIRECTORY__ has the wrong value so #load does not work correctly. Sending the same selection the second time (fsi terminal is already open) works correctly and __SOURCE_DIRECTORY__ is set to the directory of the fsx script file.

### Steps to reproduce
1. Create fsx script in a subdirectory with "__SOURCE_DIRECTORY__" as content
2. Close all fsi terminals
3. Select "__SOURCE_DIRECTORY__" in fsx script and press alt+enter (FSI: Send selection)
4. do it again and see that the two outputs are different

### Expected behaviour

Sending __SOURCE_DIRECTORY__ first and second time to fsi should generate the same outcome.

### Screenshots
image

## Machine info
 - OS: Windows 10
 - .NET SDK version:
    .NET SDKs installed:
      2.2.402 [C:\Program Files\dotnet\sdk]
      3.0.102 [C:\Program Files\dotnet\sdk]
      3.1.300 [C:\Program Files\dotnet\sdk]
      3.1.403 [C:\Program Files\dotnet\sdk]
      5.0.302 [C:\Program Files\dotnet\sdk]
 - Ionide version: v5.7.1

baronfel:
This fixes #1562.

Instead of tracking a bunch of mutables, this PR moves to a file-path-keyed map of terminal records. These records are created only in one place (`handleOpenTerminal`), and removed in only 2 places, so the lifecycle of them is much easier to manage.

The terminals that are created also have better parameters now: when we spawn a terminal it's always for the currently-opened file (or if no editor is open, `tmp.fsx` in the current workspace root), and this is used to track the instances. Because these parameters are baked into the terminal creation options, we can easily refer to them in the `handleOpenTerminal` callback to make sure our bookeeping is correct.

### DIFF
--- a/src/Components/Fsi.fs
+++ b/src/Components/Fsi.fs
@@ -233,22 +233,6 @@ module Fsi =
             let dir = workspace.rootPath.Value
             tempFsxFile dir, dir
 
-    type SendContext = { cwd: string; message: string }
-    type CDResult = NoChange of FsiTerminal | Change of FsiTerminal
-
-    let sendCd (fsiTerminal: FsiTerminal) (file, dir) =
-        let cdChanged, fsiTerminal' =
-            if fsiTerminal.CWD = dir then false, fsiTerminal
-            else
-                fsiTerminal.Terminal.sendText($"# silentCd @\"%s{dir}\";;\n", false)
-                true, { fsiTerminal with CWD = dir }
-        let fileChanged, fsiTerminal'' =
-            if fsiTerminal'.File = file then false, fsiTerminal'
-            else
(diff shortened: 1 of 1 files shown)


TheAngryByrd:
🧹 Still want to keep this PR around @baronfel?

baronfel:
no :( this would be great to have at some point, though

Issue #1568, opened 2021-08-17
Intellisense not working


chriskuech:
### Describe the bug**
The extension does not seem to work at all.  There is no error highlighting or completion recommendations.

### Steps to reproduce
Open an F# project in VS Code on macOS.

### Link to sample reproduction
N/A

### Expected behaviour
Syntax highlighting is aware of dependencies, errors are red underlined, autocompletions work etc

### Screenshots

![image](https://user-images.githubusercontent.com/3680671/129659185-afc3ce10-2fa7-4253-b578-c8314cfefe85.png)
![image](https://user-images.githubusercontent.com/3680671/129659218-e3ed3aa8-1c2d-42df-915c-f4078296e914.png)

## Machine info
 - OS: Mac
 - .NET SDK version: 5.0.400
 - Ionide version: `latest`

### Additional context
Projects work fine in Visual Studio for Mac, just not VS Code.

klettier:
Same here the extension is not working anymore, here is the VSCode log

```
console.ts:137 [Extension Host] (node:14176) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.(Use `Code --trace-deprecation ...` to show where the warning was created)
S @ console.ts:137
console.ts:137 [Extension Host] DOTNET PATH: C:\Program Files\dotnet\dotnet.exe
console.ts:137 [Extension Host] FSAC (NETCORE): error while invoking: Command failed: C:\Program Files\dotnet\dotnet.exe --version'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
console.ts:137 [Extension Host] FSAC (NETCORE): 'c:\Users\FSharp\.vscode\extensions\ionide.ionide-fsharp-5.7.1/bin/fsautocomplete.dll'
notificationsAlerts.ts:40 The F# server crashed 5 times in the last 3 minutes. The server will not be restarted.
onDidChangeNotification @ notificationsAlerts.ts:40
console.ts:137 [Extension Host] [LineLens] Activating
console.ts:137 [Extension Host] [PipelineHints] Activating
mainThreadExtensionService.ts:63 [[object Object]]Cannot read property 'indexOf' of undefined
$onExtensionRuntimeError @ mainThreadExtensionService.ts:63
mainThreadExtensionService.ts:64 TypeError: Cannot read property 'indexOf' of undefined
[...]

PR #1588, opened 2021-09-22
wrap the path to dotnet.exe when it contains spaces

baronfel:
Potential solution for some errors reported in #1568.

The process-spawning library isn't auto-escaping fully-qualified commands that contain spaces, and by default the path to `dotnet.exe` on windows is such a path.  Will need to test on a windows machine to verify this.  If this is the case, then we'll want to push this space-escaping down into our process-helper library as a follow-up action, so that we can remove this small change as well as get this kind of protection everywhere we use the library.


baronfel:
closing, implicitly done via other PRs

Issue #1591, opened 2021-09-24
Change the name in Command Palette to ionide or fsharp (better results for searching)


sharno:
VSCode mostly shows `F` results when searching for `F#` in the command palette. It also feels a lot easier to write `fsharp` or `ionide` to reach quickly for a command from the extension.

The photo shows that only the `F` is being highlighted in the higher results. Also showing that there are many results before the actual extension commands.
![image](https://user-images.githubusercontent.com/601606/134602899-2d3c96ac-39db-4d67-bc94-b2ca9df75b17.png)

baronfel:
That's not a bad idea. it wouldn't be hard either, just changing a bunch of the `contributes.commands` entries.

Krzysztof-Cieslak:
I'm not really sure if I am comfortable with this change - it's a breaking change for the users, it's a longer thing to type, and it's not as obvious as F#. I think a better solution here would be complaining about this issue to the VSCode team. :-)

sharno:
- I think actually that this is an additive change. It doesn't really break any commands that you write because of the fuzzy search. 
- Second is that because VSCode doesn't really consider the `#` sign in the search, so typing `f: run` is equal to typing `f#: run` as you see in the photo it's matching everything including the `#` sign. 
- Lastly, the `#` sign is really not natural to type while developing in F# and it needs a key combination, so I it doesn't feel really natural while exploring/using the extension to use a key combination while typing in the search bar. 

Did I manage to convince you? 😄 


![image](https://user-images.githubusercontent.com/601606/134689678-72edbfa9-aeff-42b2-a4bd-c50a6bb927d8.png)


![image](https://user-images.githubusercontent.com/601606/134689599-2edca47e-ec2c-4ce5-ad04-d8d202e4c623.png)

HerrNiklasRaab:
I agree with @sharno. I am in favor of the name "Fsharp".

Krzysztof-Cieslak:
Converting this to the discussion rather than issue.

PR #1592, opened 2021-09-24
Solves #1591: make the command palette get better results by switching command names from F# to FSharp

sharno:


Krzysztof-Cieslak:
This has been fixed on the VSCode side, and the command palette behaves as expected.
image

Issue #1107, opened 2019-05-22
Ionide no longer honoring editor.smartTabs; backspace deletes one space instead of one indent level


rmunn:
**Describe the bug**
With Ionide extension enabled, `editor.smartTabs` setting is no longer honored: backspace key is deleting one space instead of one indent level.

Disabling the ionide-fsharp extension restores normal Backspace functionality (deletes one indent level as per VS Code's normal `editor.smartTabs` behavior), and reenabling the ionide-fsharp extension brings the buggy behavior back, so I know this is caused by the extension.

I thought at first that the `FSharp.smartIndent` setting might be influencing this, but I've seen the buggy behavior whether or not `FSharp.smartIndent` was enabled.

**To Reproduce**
1. Make sure the `editor.smartTabs` setting is checked (enabled).
1. Press Backspace at the start of an indented line of code.

**Expected behavior**
One indentation level (four spaces in my case) is deleted.

**Actual behavior**
One space is deleted, and I have to press Backspace three more times to delete the indentation level.

**Environment (please complete the following information):**
 - OS: Ubuntu 18.04.2
 - Ionide version: 3.38.0
 - VSCode version: 1.34.0
 - dotnet SDK version: 2.2.300
 - mono / .Net Framework version: Mono 5.20.1.19
<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz (8 x 3449)|
|GPU Status|2d_canvas: disabled_software<br>checker_imaging: disabled_off<br>flash_3d: disabled_software<br>flash_stage3d: disabled_software<br>flash_stage3d_baseline: disabled_software<br>gpu_compositing: disabled_software<br>multiple_raster_threads: disabled_off<br>native_gpu_memory_buffers: disabled_software<br>rasterization: disabled_software<br>surface_synchronization: disabled_off<br>video_decode: disabled_software<br>webgl: disabled_off<br>webgl2: disabled_off|
|Load (avg)|1, 1, 1|
|Memory (System)|15.57GB (0.44GB free)|
|Process Argv|.|
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (9)</summary>

Extension|Author (truncated)|Version
---|---|---
EditorConfig|Edi|0.13.0
Ionide-FAKE|Ion|1.2.3
Ionide-fsharp|Ion|3.38.0
[...]

PR #1609, opened 2021-10-31
Fix #529 and #1107: Backspace should delete one indent level

### LINKED ISSUE #529: Pressing Backspace after pressing Tab erases only one space.

Tasiro:
Fixes #529 and #1107: Backspace should delete one indent level

This PR removes the F# language configuration from the `package.json`, so that the `editor.useTabStops` setting in the `settings.json` of the user will prevail.

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -1124,12 +1124,6 @@
 				"when": "editorFocus && editorLangId == 'fsharp'"
 			}
 		],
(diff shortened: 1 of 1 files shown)


baronfel:
Thank you! I'll merge this and make a release in the next day or so, depending on if I can land .net 6 support for FSAC that soon or not.

PR #1638, opened 2021-11-20
Enable Info Panel in F# scripts

### LINKED ISSUE #1625: Info Panel and info "bottom pane" not working in scripts.
### Describe the bug
The Info Panel works fine for project files, but not scripts. The same applies for the Ionide panel at the bottom of the window - it simply doesn't show anything.

### Steps to reproduce
1. Create a script.
1. Create any valid F# code.
1. Open Info Panel.
1. Move the caret over a valid symbol or keyword.
1. No content is shown in the Info Panel or the Ionide Pane.

### Expected behaviour
Both UI elements work as expected.

## Machine info
 - OS: Windows
 - .NET SDK version: 6.0
 - Ionide version: 5.9.0

Krzysztof-Cieslak:
Fix #1625

### DIFF
--- a/src/Components/InfoPanel.fs
+++ b/src/Components/InfoPanel.fs
@@ -17,7 +17,7 @@ module InfoPanel =
             if JS.isDefined textEditor && JS.isDefined textEditor.document then
                 let doc = textEditor.document
                 match doc with
-                | Document.FSharp -> true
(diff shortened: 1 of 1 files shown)


baronfel:
Yep, that'll do it.

PR #1642, opened 2021-11-24
Move the cursor down when sending a line to FSI

HoraceGonzalez:
I'm not sure if this is a known issue, but several months ago, something broke the functionality that moves the cursor down when sending a line to FSI. I believe this PR fixes this issue by executing the `cursorMove` command per the vscode documentation: https://code.visualstudio.com/api/references/commands (see screenshot below):

image


Krzysztof-Cieslak:
Thanks!

PR #1645, opened 2021-11-27
 Contribution Guide and README edits

HoraceGonzalez:
I had some finding the build & set-up instructions in the readme and thought it might help to make these visible.

I also noticed that the the contribution guide was a bit out of date and could use a little TLC. I updated the setup instructions (with a video demoing how to run Ionide in watch mode) and added a ToC to make them easier to locate. 

I hope you don't mind that I took some editorial liberties. I tried my best to rephrase so the language would be more succinct and direct without changing the meaning/intent.


Krzysztof-Cieslak:
This is amazing, thanks a lot!

PR #1646, opened 2021-11-29
Fix some indexOf errors in the FSI watcher feature

baronfel:
This resolves about half of the random indexOf errors that users
report. The `buf` in all cases is `undefined`, but when we call ToString()
on it we get the string literal `"undefined"`, which then we attempt to
process as if it was the output of the watcher.fsx script.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -19,6 +19,12 @@
       "commands": [
         "fable"
       ]
+    },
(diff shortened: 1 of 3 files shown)


Krzysztof-Cieslak:
Haven't we already merged this one #1643 ?

baronfel:
yeah, I'm bad at git

PR #1658, opened 2022-01-21
Update infrastructure

Krzysztof-Cieslak:
* Bump infrastructure to .NET 6
* Reference FSAC as a package instead of GitHub repo
* Use FAKE with console app instead of script


baronfel (approved):
LGTM. I bumped FSAC to 0.50.0 and updated the release notes as well, so when you merge this we should be ready to do another release.

PR #1671, opened 2022-03-21
Test explorer

nojaf:
I'm trying to get the NUnit tests working.
Still a lot of work that needs to be done.
I've made a change in the approach, I think it makes more sense to group all the tests underneath the individual projects.
I'm also not the biggest fan of the [testing API](https://code.visualstudio.com/api/extension-guides/testing), I found myself needing additional data every step along the way.

Anyway @baronfel, @Krzysztof-Cieslak let me know if I'm heading in the right direction.
I could use a hint in the best way to run the tests and process the output.


baronfel:
@nojaf I think there's actually a gap in the API here: we can't easily collect process/task execution status currently: https://github.com/microsoft/vscode/issues/97475

I think given that limitation it's ok to keep the current msbuild execution mechanism.

nojaf:
Hey @baronfel, I'm trying to pick up the pieces here.
Any idea how I can show the user that we are invoking MSBuild to build and later run the tests?

baronfel:
@nojaf perhaps the [progress](https://code.visualstudio.com/api/references/vscode-api#Progress) APIs would be a good mechanism? we use it in a few areas of the codebase already so you could draw inspiration from there

baronfel:
@nojaf we were already using the progress APIs in the MSBuild API you were calling, so I fleshed out that API to be a bit more descriptive. What do you think?

nojaf:
I tried the notifications instead, I found that a bit more visible.
But I would be ok to change this.
Thanks for the pointers here Chet, finally made some progress again (pun intended).

nojaf:
This is by no means a perfect implementation but might be sufficient for a first release.
All tests are run via `dotnet test`, even the Expecto ones.
This way, there is no special code for any framework. To start with this seems plausible.

baronfel:
That seems fine to me, though (note to self) we should document that on the ionide.io website when writing docs for this feature/release.

baronfel:
I spent some time with this this morning, and I definitely think it works well enough for a first release.  There are some additional enhancements that can come after (debugging tests is the big one that comes to mind), but this is already a great addition.

PR #1675, opened 2022-03-25
Display non-latin files in Solution Explorer

kant2002:
Tested with Cyrillic symbols

# Before 

image

# After

image


baronfel:
Thank you! Will try and get a patch release out later today.

PR #1676, opened 2022-03-27
Create launch configurations from launchSettings.json and project metadata

baronfel:
This looks for `Properties/launchSettings.json` in projects that exist, and makes launch/debugging configurations for those that do. This is nice because launchSettings.json can be shared across VS, VSCode, Rider, and the dotnet CLI, instead of each system needing separate files.

Notes on the design:
* These are accessed by the "Debug: Select and Start debugging" command palette option, underneath the `.NET 5+ and .NET Core` menu item
image
image
* The name of each configuration is `<project file name>: <launchSettings profile name>`
image
* only profiles of `commandName = "Project"` are handled. This is in line with `dotnet run`, but could be expanded. Other options mostly deal with IIS Express.
* if `launchBrowser` is true in the launch settings profile, the `serverReadyAction` is set to look for the server url in the classic Kestrel log message `Now listening on: <url>`, so make sure that's logged
* `environmentVariables` are pushed into the launch configuration
* If your project has an OutputType of Exe, a default launch configuration will be generated for it

Unfortunately, VSCode currently restricts compound launch configurations to those listed in launch.json, so these dynamically-generated launches are not suitable for multi-project run scenarios.

### DIFF
--- a/src/Components/Debugger.fs
+++ b/src/Components/Debugger.fs
@@ -200,6 +200,13 @@ module Debugger =
 
     /// minimal set of properties from the launchsettings json
     [<Interface>]
+    type JsMap<'t> =
+        [<Emit("$0[$1]")>]
+        abstract member Item: string -> 't
+        [<Emit("Object.keys($0)")>]
+        abstract member Keys: string []
+
+    [<Interface>]
     type LaunchSettingsConfiguration =
         abstract member commandName: string option
         abstract member commandLineArgs: string option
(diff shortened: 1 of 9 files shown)


src/Components/Debugger.fs
```diff
@@ -195,6 +198,187 @@ module Debugger =
...
+                    |> Seq.collect (fun (p: Project) ->
+                        match readSettingsForProject p with
+                        | Some launchSettings ->
+                            let projectConfigs = configsForProject (p, launchSettings)
+
+                            if Seq.isEmpty projectConfigs then
+                                (Option.toList (defaultConfigForProject p)
+                                 |> List.toSeq)
+                            else
+                                projectConfigs
```
Krzysztof-Cieslak, line 357: Should we also append the default one in case the user have `launchSettings` ?

baronfel: My thinking was something like: "having a launchSettings file means a user made an explicit choice in how the app needs to be run, so we should defer to that".

That might not always be the case, and/or the file might be used for certain specific launch _modes_, so I can see the argument for adding the default configuration too. I mostly worry about choice overload for the user, but a) we limit this to runnable project types already, b) the list is filterable by typing, and c) recently-used configurations are pinned to the top of the list, so I think that's a good UX overall.

PR #1683, opened 2022-04-01
Client side of Inlay Hint support

baronfel:
This is the VSCode-side of inlay hint support, the backend is provided at https://github.com/fsharp/FsAutoComplete/pull/907.

There's nothing really special from the VSCode side of things, all the interesting work is on the backend.

Here's a snippet of what it looks like from my local debug version of the editor (which has bad coloring, I know):

image


baronfel:
I've added config toggles for enabling the feature overall, as well as type annotations and method parameters specifically. I'll have to do a PR to the ionide docs to sync this up, but they show up in the VS Code settings UI nicely.

Issue #1672, opened 2022-03-23
Use ESBuild instead of Webpack


nojaf:
Hello, as webpack is no longer what the cool kids use, I'm wondering if you would be open to swapping it for [esbuild](https://code.visualstudio.com/api/working-with-extensions/bundling-extension#using-esbuild)?

cartermp:
I'm not opposed in theory, but my experience with js build systems is rooted in much pain and so I wonder how much effort it would be to make the change.

baronfel:
Yeah, it's just an effort question to me. The current setup has worked for a while and it's the part of the codebase I'm least-comfortable with, so I'm change-averse.

AngelMunoz:
I might give it a try this weekend, one can know if it'll work under a couple of hours so if I face anything weird I'll report back

nojaf:
Nice! Would be great if we could get a `fable` and `esbuild` watch kind of thing going on.

baronfel:
hot reload?! https://blog.hediet.de/post/hot_reload_for_vs_code_extension_development

AngelMunoz:
I went out and checked (with  vite as opposed to raw esbuild) and after some inspection and looking after the code I have a few questions:

- What is the main reason webpack is/was being used?
- Do we even want webpack/esbuild/vite at all?

I can only think webpack was there to transpile back early fable code where electron didn't support esm, and perhaps even some ES201X features, this is no longer the case the recent node/electron versions support all of the features that fable transpile as far as I'm aware, we could ditch it entirely to be honest.

The only reason I can think to have an extra build/bundling tool would be if extensions need to have bundled javascript files rather than transpiled and minified source code.

I think we should be able to just run fable in watch mode in the back and then just launch the extension with the direct output of fable there would be some work to be done in regards imported members without extensions but I think over all it could be worth exploring

thoughts?

baronfel:
That's an interesting idea for sure. I don't think we do anything special - as long as we can emit source maps that should be all we need. The VSCode docs highly suggest bundling for a perf benefit, but realistically our extension isn't really that large at all. Perhaps shipping an unbundled version would be better for debugging?

nojaf:
I would still bundle for production as the docs say, you don't know upfront what target (desktop/browser) you are dealing with so you want maximum coverage.
It would be super cool if while developing you would only need Fable (in watch mode) and can have that extension reload working. That would really be the sweet spot I think.

AngelMunoz:
[...]

PR #1684, opened 2022-04-05
Replace webpack with esbuild.

nojaf:
Fixes https://github.com/ionide/ionide-vscode-fsharp/issues/1672.
Followed the instructions at https://code.visualstudio.com/api/working-with-extensions/bundling-extension#using-esbuild.
The watch target is no longer building, I was able to run the launch target after Fable compiled the files.
What a world we live in.


MangelMaxime:
Hello,
hum in this PR @nojaf you are removing `yarn.lock` file which is not related to webpack. 

I think this is an error on your side or that your changes are incomplete as `yarn` is still used in the build script

https://github.com/ionide/ionide-vscode-fsharp/blob/69889cc1b006f89118aa4f08edd08e39132fe363/build/Program.fs#L279-L280

or even the workflow:

https://github.com/ionide/ionide-vscode-fsharp/blob/69889cc1b006f89118aa4f08edd08e39132fe363/.github/workflows/pr.yml#L30

nojaf:
Hey Maxime, good spot, I indeed did not indent to remove the yarn lock but commit the modified version.
I rectified this.

nojaf:
Aha, I see, yeah swapping this is good for the node_modules footprint.

AngelMunoz:
Hey hope you don't mind this, I re-tried to remove webpack from the toolchain in https://github.com/ionide/ionide-vscode-fsharp/pull/1722 I don't think we need esbuild (preprocess javascript) at all so I went with rollup (just bundles and tree-shakes)

in terms of bundling rollup is more flexible (but still simpler than webpack by a mile or two) and could allow us to work in some more complex webviews in the future

Feel free to let me know if that works or should we stick with esbuild

nojaf:
> can provide multiple bundles 

You could have multiple invocations of `esbuild` if you need that.

> a better fit from a configuration standpoint

Replacing the `webpack` config with a `rollup` will have similar story beats: there is a config file nobody wants to touch and there are multiple npm packages involved where updating a single one of them can be a lot of fun. I kinda liked the esbuild approach, where it is a single binary, single command and light footprint in terms of `node_modules`. From that point of view, rollup seems like a step back under the pretence that it will solve future problems which I don't see yet right here, right now.

If we are certain that the ES20XX code by Fable runs everywhere I'd even be in favour of not bundling at all. It seemed the safest thing to do, but if that is a thing of the past then there is no need.

MangelMaxime:
@nojaf Bundling still makes sense nowdays because:

- You can minify the code and so there are less stuff to access
- Remove the comments which don't need to parse at runtime
- Don't need to access severals files to load but only a single one (true it means loading everything at once instead of on the needs)

My 2 cents, is that if esbuild does the job let's use it. If in the feature we need specific features that esbuild doesn't support it is always possible to revisit the build process.

AngelMunoz:
> Replacing the webpack config with a rollup will have similar story beats: there is a config file nobody wants to touch and there are multiple npm packages involved where updating a single one of them can be a lot of fun.

The libraries surrounding rollup are fairly more stable and the packages needed are actually updated by the core rollup team, we don't need to depend on packages outside core, in any case the configuration for rollup is far simpler than that of webpack, we don't need to meddle with loaders and weird resolution rules, since rollup is meant to be used with modern javascript.

> the other libraries I _installed_ on that PR were actually undeclared dependencies (we were using them but because of the flat modules of yarn/npm it was easy to require them by any package we used)


>  I kinda liked the esbuild approach, where it is a single binary, single command and light footprint in terms of node_modules.

I agree there it's a simpler alternative and in conjunction with pnpm disk footprint can lower down quite a lot (improving CI times as well)

> From that point of view, rollup seems like a step back under the pretence that it will solve future problems which I don't see yet right here, right now.
>
> If we are certain that the ES20XX code by Fable runs everywhere I'd even be in favour of not bundling at all. It seemed the safest thing to do, but if that is a thing of the past then there is no need.

I actually asked that question https://github.com/ionide/ionide-vscode-fsharp/issues/1672#issuecomment-1079297759 _whether we wanted bundling or not_

I don't think we do, but if we need to bundle anyway because the docs says it is a good practice I personally would trust more rollup than esbuild in that aspect but that's just personal opinion

Issue #1693, opened 2022-04-20
Inline hints are too intrusive


dsyme:
I think the inline type annotations are too intrusive as things stand, it just feels like the highlighted grey boxes are the majority of what I see. I'll keep an eye on it but my gut feeling is 
* The emphasis box around the annotation is visually intrusive and takes attention away from the actual code
* We shouldn't be annotating variables bound in pattern matching by default.
* We should indicate how to toggle the annotations on/off as a hover-tip over the annotations themselves 

e.g. in this code:

![image](https://user-images.githubusercontent.com/7204669/164155104-77540502-1aee-4e1f-9316-6a9bc5df052f.png)

or this code

![image](https://user-images.githubusercontent.com/7204669/164155149-9cb6f364-f747-4a3b-9dfb-b61321451b1e.png)

baronfel:
This has been a common request, but it is documented already. What we could add is mention of the relevant config options directly in the docs about inlay hints, instead of only in the full config options listing.

Separately, there are too many hints reported at the moment, type hints for already typed expressions are unnecessary and will be fixed in the next release.

Finally, vscode itself will be changing the default styling for these hints to a much more subdued default in the next version, which should help immensely.

dsyme:
Adding parameter and union-case names by default looks like too much - a List.map is now hard to read:

![image](https://user-images.githubusercontent.com/7204669/164156174-36b2026b-0e4f-43a1-bf13-c914baca8026.png)

dsyme:
Note I'm really only concerned about the defaults - I don't configure things like this, the defaults have to be right.

forki:
I personally went with type hints on and parameter name hints off. Looks
much quieter.

cartermp:
I expect there to be a few growing pains here. Fundamentally, I think inlay hints are more powerful than tooltips. You get the type definitions for a whole viewport at once, which can really help in many situations, such as a beginner accidentally forgetting to apply all parameters to a function call and having that impact stuff downstream in unexpected ways. That kind of scenario just wasn't dealt with well prior to inlay hints. And since they're just little blocks in the editor, you can interact with them (pop a full tooltip, use it to add a type annotation, etc).

In Visual Studio, they're off by default in C# and need a toggle (alt+f1) which all but kills them as a feature because they're hidden by an obscure key command. But the baby need not be thrown out with the bathwater - I think the right model is actually toggling them on when you need them.

Is there a VSCode story for toggling them on/off?

[...]

PR #1695, opened 2022-04-22
Nicer UX for inlay hints

baronfel:
Part of #1693 

The main thing this does is overhaul the tooltips for the inlay hints - now there are command-links (clickable links) that will disable:
* this specific kind of hint (type hint vs parameter name)
* all hints
* (if the version of VSCode is at least 1.67.0) a configuration that lets the user press Ctrl+Alt to toggle the display of the hints while held
* dismissing the clickable tooltips here

This was actually pretty easy to set up, we should do things like this more often!

### DIFF
--- a/src/Components/InlayHints.fs
+++ b/src/Components/InlayHints.fs
@@ -9,19 +9,36 @@ open Fable.Core.JsInterop
 let private logger =
     ConsoleAndOutputChannelLogger(Some "InlayHints", Level.DEBUG, None, Some Level.DEBUG)
 
-let enabled () : bool =
-    Configuration.getUnsafe "FSharp.inlayHints.enabled"
+let mutable private toggleSupported = false
+module Config =
+    let enabled = "FSharp.inlayHints.enabled"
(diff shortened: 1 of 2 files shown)


src/Components/InlayHints.fs
```diff
@@ -9,31 +9,73 @@ open Fable.Core.JsInterop
...
+        if useLongTooltip
+        then
+            let lines = ResizeArray()
+            let hideCommand =
+                match kind with
+                | "Type" -> Commands.hideTypeAnnotations
+                | "Parameter" -> Commands.hideParameterNames
+                | _ -> ""
+            lines.Add $"To hide these hints, [click here](command:{hideCommand})."
+            lines.Add $"To hide _all_ hints, [click here](command:{Commands.hideAll})."
```
cartermp, line 71:
```suggestion
            lines.Add $"To hide *ALL* hints, [click here](command:{Commands.hideAll})."
```
I think going all-caps is good here, more prominent the better

cartermp (approved):

Krzysztof-Cieslak:
Lgtm

PR #1700, opened 2022-04-30
Bump FSAC and absorb new config entries and API shapes

baronfel:
@Krzysztof-Cieslak this is working great for me, wanna give it a spin before release?


baronfel:
This looks good! @Booksbaum's work on the inlay hints especially really declutters things.

PR #1713, opened 2022-05-23
Update to vscode-languageclient 8

baronfel:
This required mostly just changing the shape of a documentselector (since that was changed in VSCode 3.16 but we never adapted to it). Along the way debugging that, though, I got frustrated with webpack's sourcemap-based debugging not working and ripped it out entirely, replacing it with esbuild (based in part on #1684 but replacing every part of the build).

I also updated to latest fantomas.  I _believe_ this is working in its entirety, but before merge I want to:

* [x] have a PR to the languageclient F# types to fix the DocumentSelector bindings
* [ ] ensure f5 debugging is still good here (fake integration and whatnot)
* [ ] test the built vsix locally


src/Components/CodeLensHelpers.fs
```diff
@@ -4,13 +4,10 @@ open System
 open Fable.Core.JsInterop
 open Fable.Import.VSCode
 open Fable.Import.VSCode.Vscode
-open global.Node
-
-module node = Node.Api
```
baronfel, line 9: these kinds of module node declarations were removed globally because of F# language warnings about partially-resolved namespace names.

src/Core/LanguageService.fs
```diff
@@ -626,7 +631,9 @@ Consider:
             synch.configurationSection <- Some !^ "FSharp"
             synch.fileEvents <- Some(!^ ResizeArray([ fileDeletedWatcher ]))
 
-            opts.documentSelector <- Some !^selector
+            // this type needs to be updated on the bindings - DocumentSelector is a (string|DocumentFilter) [] now only.
```
baronfel, line 634: this is the actual fix, will PR to the type definitions

src/Core/Utils.fs
```diff
@@ -421,11 +420,11 @@ type ShowStatus private (panel: WebviewPanel, body: string) as this =
 module VSCodeExtension =
 
     let private extensionName =
-        #if IONIDE_EXPERIMENTAL
-
+#if IONIDE_EXPERIMENTAL
+        "ionide-fsharp-experimental"
```
baronfel, line 424: had to fix this missing string or else fantomas got real sad

src/Core/LanguageService.fs
baronfel: Done now.

baronfel:
Something's very off here - LSP features don't seem to be working at all.

baronfel:
probably going to revert most of this in favor just of the documentselector change - too much change is making it hard to pin down where errors are coming from

baronfel:
@AngelMunoz I could use some help here if you have time.

The new versions of vscode-languageclient apparently use syntax that the current webpack configuration doesn't understand. When I bundle, I get the following output:

```
ERROR in ./node_modules/vscode-languageclient/lib/node/main.js 267:50
Module parse failed: Unexpected token (267:50)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|                 if (node.runtime) {
|                     const args = [];
>                     const options = node.options ?? Object.create(null);
|                     if (options.execArgv) {
|                         options.execArgv.forEach(element => args.push(element));
 @ ./out/Core/LanguageService.js 31:0-55 1059:15-29
 @ ./out/fsharp.js
```

To me, this means that I need to preprocess this dependency somehow? So I attempted to explicitly include all of the vscode dependencies in the bundling, but that didn't solve it. Do you have any ideas?

I'd like to not take on the esbuild transition in this work, because I've tried it and ended up with the extension not working at all, so I'm a little frazzled :(

AngelMunoz:
I'll try to give it a check tomorrow but this looks odd, is the language client available in vscode itself at runtime? you could mark it as an external in the webpack config (just like vscode is an external) if that's the case, otherwise we might want to take this babel plugin to fix this particular issue 
https://babeljs.io/docs/en/babel-plugin-proposal-nullish-coalescing-operator

baronfel:
ok, we don't need babel at all in newer versions of webpack! I've updated us and we're all set as far as I can tell.

Krzysztof-Cieslak (approved):
LGTM

Issue #1711, opened 2022-05-15
FSI only shows "Error: An index satisfying the predicate was not found in the collection." for local values


jcmrva:
### Description
When an FSI session is started by sending code from a file (fs or fsx), values are replaced with "Error: An index satisfying the predicate was not found in the collection."

### Steps to reproduce
1. Select a complete line of code.
2. Use `Alt+Enter` or `Send Selection` in the command palette.

> example.fs

let x = 1

> fsi

```
> let x = 1;;     
val x: int =
  Error: An index satisfying the predicate was not found in the collection.
```

### Expected behaviour
Value appears as expected, e.g.
```
> let x = 1;;     
val x: int = 1
```

## Machine info
 - OS: Windows
 - .NET SDK version: 6.0.300
 - Ionide version: 6.0.4

### Additional context

The most recent change was updating to `SDK 6.0.300` so I suspect that's the culprit. I have `SDK 7 preview 4` on another machine and the error does not occur there.

I'm not able to reproduce this in VS, Rider, or an FSI session started with `dotnet fsi` (terminal or within VS Code).

baronfel:
I repro on Windows, would be interested in seeing if it repros on MaOS or Linux

jcmrva:
Pretty sure I've identified the cause, going to see if I can fix it.

baronfel:
Heck yeah! Let me know if you need any guidance.

jcmrva:
Quick fix: turn off FSI Watcher and restart VS Code. (the setting doesn't say to restart but it appears necessary)

It bombs on this line:
https://github.com/ionide/ionide-vscode-fsharp/blob/6c0a9e7e34d279f9dd6541db421b30501a3231b4/release/watcher/watcher.fsx#L4

...because there are tons of them:
>    FSI-ASSEMBLY1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null;
>    FSI-ASSEMBLY2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null;
>    FSI-ASSEMBLY3, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null;
>    FSI-ASSEMBLY4, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null;
>    FSI-ASSEMBLY5, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null;
[...]

PR #1716, opened 2022-05-27
check all FSI assemblies

jcmrva:
For #1711 

This isn't doing everything I intended, most notably handling the shadowed declarations, but I thought I should get a PR up.

I did manage to maintain compatibility with the single assembly.

### DIFF
--- a/release/watcher/watcher.fsx
+++ b/release/watcher/watcher.fsx
@@ -94,7 +94,15 @@ fsi.AddPrinter (fun (_: obj) ->
     let formatVarsAndFuncs name value typ step =
         (sprintf "%s###IONIDESEP###%A###IONIDESEP###%s###IONIDESEP###%i" name value typ step).Replace("\n", ";")
 
+    let formatVarsAndFuncs' name value typ step (lookup:Set<string>) =
+        let name = if lookup.Contains name then (name + " (shadowed)") else name
+        (sprintf "%s###IONIDESEP###%A###IONIDESEP###%s###IONIDESEP###%i" name value typ step).Replace("\n", ";")
+
     let formatRecsAndUnions name flds step =
+        (sprintf "%s###IONIDESEP###%s###IONIDESEP###%i" name flds step).Replace("\n", ";")
+
+    let formatRecsAndUnions' name flds step (lookup:Set<string>) =
+        let name = if lookup.Contains name then (name + " (shadowed)") else name
         (sprintf "%s###IONIDESEP###%s###IONIDESEP###%i" name flds step).Replace("\n", ";")
 
     let arrangeVars fsiAssembly =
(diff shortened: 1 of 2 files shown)


jcmrva:
Maybe I'm losing it. When `folder` in `allVars` is run independently, it gives the expected results. When given to List.fold, the set is updated but the list remains empty. Do I need a different approach here?

jcmrva:
This might be ready to go, aside from the copypasta in the "arrange" functions, which I will clean up at some point.

I noticed `fsi.AddPrinter` is invoked for *each value printed in the REPL*, so...
1. there probably isn't much optimization we can do within it.
2. performance doesn't seem to suffer horribly in spite of that, although I haven't tested huge sets of data.
3. docs should probably come with a warning.

In general, I think we can make it work better by moving some of the code into a better place (FSAC?)

baronfel:
Thank you for your work here! I'll take it for a test drive and roll it into a release in the next day or so.

jcmrva:
I thought it might be a good idea to document my workflow for the change because it's a little unusual vs standard compile-and-run projects.

Background: at a high level, the watcher panel gets values via this process:
1. the `watcher.fsx` file in the Ionide install is loaded into FSI on startup.
2. the watcher function is invoked whenever FSI prints values to the console & saves whatever it extracts from the assembly in `vars.txt`, `funcs.txt` & `types.txt` in the same subdirectory as watcher.fsx.
3. the watcher panel in Ionide refreshes itself from those files, often with a short delay.

The pleasant side-effect of this design is that I could work on the installed watcher.fsx (rather than the Ionide source) and rapidly test changes by only relaunching FSI. No need to retype example code or rebuild Ionide.

I broke the file content formatting early in the refactor, so being able to view the files directly made the feedback cycle even faster.

image

Extending this method to other areas for quick updates & user customization might be worthwhile.

baronfel:
@Krzysztof-Cieslak and I have chatted sometimes about moving from the 'just send/receive from a `dotnet fsi` process model' that we currently use, to having FSAC manage the persistent FsiEvaluationSessions the same way that VS and Rider do to power their 'interactive windows'. That would give us much nicer programmatic access to the bound values and open the door to doing more structured things. So if that's something you're interested in longer term we'd love to have your input.

regarding this PR though, let's ship it!

jcmrva:
Ah, the same method described [here](https://fsharp.github.io/fsharp-compiler-docs/fcs/interactive.html)?

baronfel:
yep, exactly. there would be some book-keeping around identifying _which_ evaluation session a particular interaction command would be associated with, so we'd want to spec out a small set of custom LSP endpoints specifically for dealing with this, but it would be possible to do something like:

* User sends line/selection/file to FSI session via custom LSP endpoint
* FSAC identifies the session by some kind of ID and performs the evaluation of the provided line/selection/file
* FSAC sends any output of the evaluation to VSCode for display
* FSAC extracts the session state after the evaluation using the `session.GetBoundValues()` API and returns that data to VSCode for display in a watcher (if open)

PR #1717, opened 2022-05-29
Fix: Several commands don't work in untitled docs

Booksbaum:
Several commands don't work in untitled docs because they use FileName instead of URI:  
F1 Help, Info Panel, Quick Info, Line Lense, Pipeline Hints

These are not just Ionide fixes, but require changes in FSAC too: fsharp/FsAutoComplete#949

<br/>

Additional fix:
* No Quick Info in Script file

### DIFF
--- a/src/Components/FSharpLiterate.fs
+++ b/src/Components/FSharpLiterate.fs
@@ -216,7 +216,7 @@ module FSharpLiterate =
         let update (textEditor: TextEditor) =
             promise {
                 let doc = textEditor.document
-                let! res = LanguageService.fsharpLiterate doc.fileName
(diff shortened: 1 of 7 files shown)


baronfel:
These are great - there's a sort of long-standing problem that you're touching on her where many of the custom endpoints use non-LSP structures and as a result are prone to this sort of error. I'll take a detailed look today!

baronfel:
I've tested this locally this morning and both sets of changes are good. Thank you (yet again!) for smoothing out the rough corners of these projects.  I'll merge both and get a release with these fixes out today!

Getting contributions like this is _incredibly_ motivating for me personally, thank you for taking the time to make FSAC and Ionide better :)

Issue #1720, opened 2022-06-10
Consider managing FSI sessions in FSAC directly


baronfel:
Managing sessions directly in FSAC would allow for more full featured treatments for features like the watcher, as well as serving as the foundation for notebook approaches.

There would be some book-keeping around identifying _which_ evaluation session a particular interaction command would be associated with, so we'd want to spec out a small set of custom LSP endpoints specifically for dealing with this, but it would be possible to do something like:

* User sends line/selection/file to FSI session via custom LSP endpoint
* FSAC identifies the session by some kind of ID and performs the evaluation of the provided line/selection/file
* FSAC sends any output of the evaluation to VSCode for display
* FSAC extracts the session state after the evaluation using the `session.GetBoundValues()` API and returns that data to VSCode for display in a watcher (if open)

_Originally posted by @baronfel in https://github.com/ionide/ionide-vscode-fsharp/pull/1716#issuecomment-1150607217_

Krzysztof-Cieslak:
> User sends line/selection/file to FSI session via custom LSP endpoint

I do wonder if there's some way of just redirecting stdio into the FCS-hosted FSI session so we keep the ability to type into the terminal. 
This probably would mean that our "enhanced FSI" would be a separate process (i.e. not FSAC) but since it'd be used just for FSI-related things it shouldn't be too perf-expensive

Krzysztof-Cieslak:
Proof of concept for the idea of just wrapping FCS-hosted FSI session into console app, it also starts HTTP server which can be used for getting the info about the session: https://github.com/ionide/Falcon/blob/master/src/Falcon/Program.fs

Krzysztof-Cieslak:
CC: @jcmrva, it may be interesting for you

AngelMunoz:
This might be kind of related, I've been looking today into the FSI watcher and I found out it is very annoying to create webviews, would there be interest to also host FSI related Webviews in that out or process server?
I'm not too familiar with the code extensions API so I'm not sure if you can just use any URL but that would also give us an edge to use F# tooling for the website rather than just try to force strings for webviews

jcmrva:
@Krzysztof-Cieslak I like it!

> I do wonder if there's some way of just redirecting stdio of the console app into the FCS-hosted FSI session so we keep the ability to type into the terminal.

That would be ideal.

[...]

PR #1721, opened 2022-06-12
Add VSC UI Toolkit

AngelMunoz:
This PR attempts to ease a little bit working with webviews and to also enable the usage of the vscode ui toolkit.

I didn't go as far as thinking what else can we use webviews for but I at least switched the table for the watcher for the datagrid included in the toolkit.

There are a few observations I have on this regard

- I had to include the toolkit js file on the release directory
  
  This might be able to be solved with a few tweaks on the webpack config but I couldn't figure it out

- I tried using Feliz.ViewEngine but I decided to stick with strings
  
  The vscode toolkit would require the proper bindings for Feliz.ViewEngine and that might be quite a lot of work, thankfully Alfonso has already make it easier to work with html/css/js strings within F# files thanks to his extension so I felt it was actually easier to work with inline strings

- Webviews could be made with an project outside src

  After my comment on #1720 I found out that yeah we could use external websites for the webviews but they have to live in iframes which make things quite complicated, what what we could do is to have a separate project (Fable.Lit or Feliz based one) that outputs the final bundle to the release directory, later on we would build the index by hand (as we're doing it now) but including the scripts within the release directory. That complicates infrastructure though so it might be just too much for the project to handle at least right now.

In any case Feel free to close this PR if this is not needed it was more to explore approaches and see what worked and what not, I thought it could be useful so I ventured to open this.


baronfel:
Haven't looked at this in detail, but I am looking at moving to esbuild in #1684/#1713 so maybe that would be easier?

.gitignore
```diff
@@ -82,4 +82,7 @@ fsacpath.txt
 /out/
 
 build/bin
-build/obj
\ No newline at end of file
+build/obj
+
+# include vscode ui toolkit minified release file
+!release/toolkit.min.js
```
baronfel, line 88: super weird - can you add not a reference to the toolkit and let webpack pull it in (I'm just skimming their [getting started](https://github.com/microsoft/vscode-webview-ui-toolkit#getting-started) docs)

baronfel (commented):
Thanks! I like the contribution, its a nice styling and UX enhancement for sure. I'd like to see if the packaging issue can be fixed - ideally we'd be able to NPM reference the toolkit and let the build take care of bundling the output.

AngelMunoz:
give me a chance to look over this week/next weekend to remove webpack from the toolchain
technically we don't need it anymore the electron version handles all of the JS Fable emits, so we would likely need something like [rollup](https://rollupjs.org/guide/en/) which helps us with minification and tree-shaking, watch mode and other helpful stuff

I don't really think we need vite/esbuild because we don't really run in a browser context we're basically just doing node stuff under the hood

.gitignore
AngelMunoz: I tried following the getting started from the toolkit but their getting started assumes the directory where you launch your extension is the root of the project (hence why it had node_modules available), in our case the _release_ directory is the root so we don't have access to node_modules, we'd likely need to do some webpack magic to pull this script at dev/build which was the part I couldn't figure out

baronfel: Gotcha. I'd also be ok longer term with removing the release split. It's just confusing IMO.

Krzysztof-Cieslak:
I like this, let's merge it.

PR #1723, opened 2022-06-14
improve send selection

yuuuxt:
Thanks for creating this extension!

I want `fsi:send selection` to send the complete lines selected, instead of selection only. It's easy to have an incomplete selection (e.g. omitting `o` of first-line `open System`) using mouse. So I made this naive PR.

Now we can just briefly select the lines (don't need to be complete) to send the complete lines.

It works like this:
temp

### DIFF
--- a/release/CHANGELOG.md
+++ b/release/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 7.21.2 - 21.09.2024
+
+* FIXED: [Fix Find References in CodeLens](https://github.com/ionide/ionide-vscode-fsharp/pull/2042) from @PaigeM80
+
 ### 7.21.1 - 07.09.2024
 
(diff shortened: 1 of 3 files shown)


jasiozet:
I think that the idea of improving selection is a **great** one, but worth noting that with this Ionide looses a feature of selecting something mid line and evaluating it. I'm not sure if it is a big deal.

```functionWithTwoArguments (anotherFunction arg) (anotherFunction2 arg2)```
I could imagine that I want to select one of the parens and evaluate the middle, with that change it would always evaluate the whole line.

My idea of solving this problem, that I want to explore is to provide a shortcut for evaluating paragraphs.

```
let s = 0

let a = 1
let b = 1
printfn "%d" <| a + b

let c = 1
```

Placing a cursor anywhere in the middle section and evaluating would send the:

```let a = 1
let b = 1
printfn "%d" <| a + b
```
To REPL

yuuuxt:
@jasiozet Yes, it is better to have a new shortcut to behave like this and keep the original one unchanged.

For me the original behavior is not useful and modifying it makes minimal effort, so I did it like this and started using it.

Your execute-paragraph idea is one more step further :)

I also created #1728, maybe we can let discussions happen there?

NeoGlitchBob (approved):

PR #1731, opened 2022-07-11
Fix Reference CodeLenses

Krzysztof-Cieslak:


baronfel (commented):
sheesh! I guess we need a bunch of overloads for registercommand, eh?

PR #1733, opened 2022-07-15
Improve error message if dotnet is not found

### LINKED ISSUE #1709: If dotnetRoot is set but dotnet program doesn't exist at that location no error is displayed to the user
Hello,

it happened that I had `FSharp.dotnetRoot` set to an invalid location and was experiencing a silent error from Ionide.

### Steps to reproduce

1. Set `FSharp.dotnetRoot` to `fake_path_to_dotnet`
2. Open an directory containing an F# project
3. See that Ionide doesn't start and doesn't report any error

Additionally, you can see a log from Ionide:

`[Extension Host] DOTNET PATH: fake_path_to_dotnet\dotnet.exe`

I can have a look at how to improve `dotnet` program existence but I am not sure where or how we want to do it.

Do we want to check for the program existence here:

https://github.com/ionide/ionide-vscode-fsharp/blob/6993276a154c764499f5d56adbf4050baace77d7/src/Core/LanguageService.fs#L801-L809

I don't think we can do it here the Environment file because otherwise it will just try to find `dotnet` anywhere in the system:

https://github.com/ionide/ionide-vscode-fsharp/blob/6993276a154c764499f5d56adbf4050baace77d7/src/Core/Environment.fs#L50-L57

```fs
let dotnet =
    Configuration.tryGet "FSharp.dotnetRoot"
    |> Option.map (fun root ->
        promise {
            let dotnetPath = root </> (if isWin then "dotnet.exe" else "dotnet")

            if exist dotnetPath then // Pseudo code
                return Some dotnetPath
            else 
                return None
        }
    )
    // Because we return none if the dotnetPath doesn't exist, we will default to `tryGetTool "dotnet"` 
    // Which is probably not what the user want if he setted `FSharp.dotnetRoot` setting
    |> Option.defaultWith (fun () -> tryGetTool "dotnet")
```

## Machine info
 - OS: Windows
 - .NET SDK version: 

```
4.1.418 [C:\Program Files\dotnet\sdk]
5.0.407 [C:\Program Files\dotnet\sdk]
6.0.101 [C:\Program Files\dotnet\sdk]
6.0.104 [C:\Program Files\dotnet\sdk]
6.0.200 [C:\Program Files\dotnet\sdk]
6.0.202 [C:\Program Files\dotnet\sdk]
```

 - Ionide version: 6.0.3

MangelMaxime:
Hello,

this PR fix #1709.

~**:warning: Important:** I had to disable Fantomas formatting because it generates invalid F# code. I opened a bug about it https://github.com/fsprojects/fantomas/issues/2366~

## Changes:

- Rename `LanguageService.dotnet` to `LanguageService.tryFindDotnet`
- Improve `LanguageService.tryFindDotnet` to gives better errors message if dotnet is not found
- Remove `Environement.dotnet` and use `LanguageService.tryFindDotnet` instead
- Use a modal instead of the notification to display the error, this is because we cannot use multine in the notification making the error message not really readable
- New `dotnet` detection logic
- Return error message which differ depending on the context of the error. Did it occur because `Fsharp.dotnetRoot` is set or not.

**Important**

~Please test on Windows, if the detection works when `FSharp.dotnetRoot` is set to a valid location. In order, to detect if the file exist at that location I am calling `Environment.tryGetTool` with the full computed path. On Linux, it executes `which /home/mmangel/.dotnet/dotnet` which return OK if found, but I don't know if the Windows equivalent was with full path too.~

Test no needed anymore, I am now using the standard Note API to check if the file exist.

### Generalise calling `windows.showErrorMessage` ?

Should we call `windows.showErrorMessage` everywhere where we handle an error result from calling `LanguageService.tryFindDotnet`? If yes, should we just call it directly from `LanguageService.tryFindDotnet` and make it return the dotnet path if found and `failwith ...` if not?

### What do to with `LanguageService.compilerLocation`?

What does `LanguageService.compilerLocation` do?

In the old code if `dotnet` is not found, then Ionide try to use the `compilerLocation` which comes from the (LSP?) client. But how can the LSP client be running if `dotnet` is not found ?

If my understanding is correct, shouldn't we just remove it? --> Need to remove the commented code

If not, we need to re-add that portion of the logic. Waiting for some explanation to understand how to wire it.

## Error preview

image

### DIFF
--- a/build/Program.fs
+++ b/build/Program.fs
@@ -386,10 +386,8 @@ let buildTargetTree () =
     "YarnInstall" ==>! "Build"
     "DotNetRestore" ==>! "Build"
 
-    // Disable Fantomas because it generates invalid code F# code
-    // "Format"
-    // ==> "Build"
-    "Build"
+    "Format"
+    ==> "Build"
     ==> "SetVersion"
     ==> "BuildPackage"
     ==> "ReleaseGitHub"
--- a/build/Program.fs
+++ b/build/Program.fs
@@ -1,4 +1,4 @@
-﻿open System
+open System
 open System.IO
 open Fake.Core
 open Fake.JavaScript
@@ -386,8 +386,10 @@ let buildTargetTree () =
     "YarnInstall" ==>! "Build"
     "DotNetRestore" ==>! "Build"
 
-    "Format"
-    ==> "Build"
+    // Disable Fantomas because it generates invalid code F# code
+    // "Format"
+    // ==> "Build"
+    "Build"
     ==> "SetVersion"
     ==> "BuildPackage"
     ==> "ReleaseGitHub"
--- a/src/Core/LanguageService.fs
+++ b/src/Core/LanguageService.fs
@@ -163,24 +163,6 @@ Consider:
                     return Core.Error msg
         }
 
-// Old implementation: How does compilerLocation work?
-// If dotnet is not found on the system, client cannot be run, no?
-//                let! dotnet = Environment.dotnet
-
-//                match dotnet with
-//                | None ->
-//                    let! location = compilerLocation ()
-//
-//                    match location.Data.SdkRoot with
-//                    | Some root ->
-//                        if Environment.isWin then
-//                            return Some(path.join (root, "dotnet.exe"))
-//                        else
-//                            return Some(path.join (root, "dotnet"))
-//                    | None -> return None
-//                | Some location -> return Some location
-
-
     /// runs `dotnet --version` in the current rootPath to determine the resolved sdk version from the global.json file.
     let runtimeVersion () =
         promise {
--- a/src/Core/LanguageService.fs
+++ b/src/Core/LanguageService.fs
@@ -102,14 +102,96 @@ module LanguageService =
         else
             (fn + ".fsx")
 
+
+    let compilerLocation () =
+        match client with
+        | None -> Promise.empty
+        | Some cl ->
+            cl.sendRequest ("fsharp/compilerLocation", null)
+            |> Promise.map (fun (res: Types.PlainNotification) ->
+                let r = res.content |> ofJson<CompilerLocationResult>
+                r)
+
+    let tryFindDotnet () =
+        promise {
+            // User has specified a custom dotnet location
+            match Configuration.tryGet "FSharp.dotnetRoot" with
+            | Some dotnetRoot ->
+                // Choose the right program name depending on the OS
+                let program =
+                    if Environment.isWin then "dotnet.exe" else "dotnet"
+
+                // Compute the full path
+                let dotnetFullPath =
+                    path.join(dotnetRoot, program)
+
+                // Check if the program exists at the computed location
+                match! Environment.tryGetTool dotnetFullPath with
+                // Everything is fine, return the path
+                | Some dotnet ->
+                    return Ok dotnet
+
+                // The program does not exist, return the error
+                | None ->
+                    let msg =
+                        // The special syntax: %s{"\n" + dotnetFullPath}
+                        // Force a new line to be added, if I put %s{dotnetFullPath}
+                        // at the beginning of the next line, there is a compiler error...
+                        $"""
+Could not find `dotnet` in the `dotnetRoot` directory: %s{"\n" + dotnetFullPath}
+
+Consider:
+- updating the `FSharp.dotnetRoot` settings key to a directory with a `dotnet` binary
+- removing the `FSharp.dotnetRoot` settings key and :
+    - including `dotnet` in your PATH
+    - installing .NET Core"""
+
+                    return Core.Error msg
+
+            // No custom location was specified, try to find it from the PATH
+            | None ->
+                match! Environment.tryGetTool "dotnet" with
+                | Some dotnetFullPath ->
+                    return Ok dotnetFullPath
+
+                | None ->
+                    let msg = """Could not find `dotnet` in the path.
+
+Consider:
+- setting the `FSharp.dotnetRoot` settings key to a directory with a `dotnet` binary
+- including `dotnet` in your PATH
+- installing .NET Core
+"""
+
+                    return Core.Error msg
+        }
+
+// Old implementation: How does compilerLocation work?
+// If dotnet is not found on the system, client cannot be run, no?
+//                let! dotnet = Environment.dotnet
+
+//                match dotnet with
+//                | None ->
+//                    let! location = compilerLocation ()
+//
+//                    match location.Data.SdkRoot with
+//                    | Some root ->
+//                        if Environment.isWin then
+//                            return Some(path.join (root, "dotnet.exe"))
+//                        else
+//                            return Some(path.join (root, "dotnet"))
+//                    | None -> return None
+//                | Some location -> return Some location
+
+
     /// runs `dotnet --version` in the current rootPath to determine the resolved sdk version from the global.json file.
     let runtimeVersion () =
         promise {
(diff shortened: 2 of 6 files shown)


Booksbaum:
doesn't work:  
dotnetRoot


---------------

`where` doesn't work with a full path:
```
> where.exe "C:\Program Files\dotnet\dotnet.exe"
ERROR: Invalid pattern is specified in "path:pattern".
```
Requires instead:
```
> where.exe "C:\Program Files\dotnet:dotnet.exe"
C:\Program Files\dotnet\dotnet.exe
```

<details>
<summary>`where.exe /?`</summary>

```
WHERE [/R dir] [/Q] [/F] [/T] pattern...

Description:
    Displays the location of files that match the search pattern.
    By default, the search is done along the current directory and
    in the paths specified by the PATH environment variable.

Parameter List:
    /R       Recursively searches and displays the files that match the
             given pattern starting from the specified directory.

    /Q       Returns only the exit code, without displaying the list
             of matched files. (Quiet mode)

    /F       Displays the matched filename in double quotes.

    /T       Displays the file size, last modified date and time for all
             matched files.

    pattern  Specifies the search pattern for the files to match.
             Wildcards * and ? can be used in the pattern. The
             "$env:pattern" and "path:pattern" formats can also be
             specified, where "env" is an environment variable and
             the search is done in the specified paths of the "env"
             environment variable. These formats should not be used
             with /R. The search is also done by appending the
             extensions of the PATHEXT variable to the pattern.

     /?      Displays this help message.

  NOTE: The tool returns an error level of 0 if the search is
        successful, of 1 if the search is unsuccessful and
        of 2 for failures or errors.

Examples:
    WHERE /?
    WHERE myfilename1 myfile????.*
    WHERE $windir:*.*
    WHERE /R c:\windows *.exe *.dll *.bat
    WHERE /Q ??.???
    WHERE "c:\windows;c:\windows\system32:*.dll"
    WHERE /F /T *.dll
```

</details>

(Note: I'm using `where.exe` instead of `where` because I'm using powershell. And `where` is an alias for `Where-Object`)


So I suggest to keep directory separate:
```fsharp
let tryGetTool dir toolName =
    if isWin then
        let pattern =
            match dir with
            | Some dir ->
                $"{dir}:{toolName}"
            | None -> toolName
        spawnAndGetTrimmedOutput "where.exe" (ResizeArray [ pattern ])
        |> Promise.map (fun (err, path, errs) -> if path <> "" then Some path else None)
        |> Promise.map (
            Option.bind (fun paths ->
                paths.Split('\n')
                |> Array.map (String.trim)
                |> Array.tryHead)
        )
    else
        let tool =
            match dir with
            | Some dir -> path.join (dir, toolName)
            | None -> toolName
        spawnAndGetTrimmedOutput "which" (ResizeArray [ tool ])
        |> Promise.map (fun (err, path, errs) -> if path <> "" then Some path else None)
```
(also: no need to use `cmd`: `where` is an executable itself -> can be started directly)

MangelMaxime:
@Booksbaum Thank you for checking.

I think I will go with the less fancy way and just use `file.exist` from Node API as it should be really cross platform. And it also convey the meaning of the check we are doing at this point in the code which is checking for the file existence.

This was actually my original idea.

MangelMaxime:
I made the changes to use standard Node API to check for the file existence.

build/Program.fs
```diff
@@ -400,8 +400,9 @@ let buildTargetTree () =
     "YarnInstall" ==>! "RunScript"
     "DotNetRestore" ==>! "RunScript"
 
-    "Clean" ==> "Format" ==> "RunScript"
-    ==>! "Default"
+    // Disable Fantomas because it generates invalid code F# code
+    // "Clean" ==> "Format" ==> "RunScript" ==>! "Default"
```
nojaf, line 404: You shouldn't disable the entirety of formatting because of one bug.
Please restore this and create a [.fantomasignore](https://github.com/fsprojects/fantomas/blob/master/docs-old/Documentation.md#ignore-files-fantomasignore) and include that one file where you have a problem.

src/Core/LanguageService.fs
```diff
@@ -115,14 +115,93 @@ module LanguageService =
...
+- setting the `FSharp.dotnetRoot` settings key to a directory with a `dotnet` binary
+- including `dotnet` in your PATH
+- installing .NET Core
+"""
+
+                    return Core.Error msg
+        }
+
+// Old implementation: How does compilerLocation work?
+// If dotnet is not found on the system, client cannot be run, no?
```
baronfel, line 180: excellent question :D compilerlocation is something that existed in the old FSAC STDIO implementation that worked for non-dotnet-SDK builds. Since we've dropped those we can kill compilerlocation entirely, both here and in FSAC.

build/Program.fs
MangelMaxime: Well, if the file is not in the last commit then build process works.

Because, Ionide doesn't run the check on all the files but only the modified ones (from HEAD).

That's kind of a strange decision. I opened an issue to discuss it #1736

src/Core/LanguageService.fs
MangelMaxime: Ok, I removed that code.

MangelMaxime:
@baronfel This PR is ready to merge depending on the decision about:

> Generalise calling windows.showErrorMessage ?

baronfel:
I worry about the user experience of having tons of error modals fired. Is there a way we could detect not finding dotnet (until the configuration changes) and then try again? Ideally notifying the user via status bar or something that we're in a degraded state?

MangelMaxime:
Another ideas, I had this morning is:

Display a minimal notifications just saying "Could not find 'dotnet'" and had an action to the notification saying "More details".

When the user click on it, we open a preview panel with the detailed error message in it.

https://user-images.githubusercontent.com/4760796/179700971-a82df49a-d017-452d-a654-def0ca7dffc5.mp4

If we go this way, I think it makes sense to "generalise calling `windows.showErrorMessage`" in the `tryFindDotnet` function as it involve a bit of logic.

```fs

let dotnetNotFound (msg: string) =
    promise {
        let! selection = window.showErrorMessage ("Could not find 'dotnet'", "More details")

        match selection with
        | Some "More details" ->
            // Prepare a temporary file to store the error message
            let fileName = "ionide_dotnet_not_found_details_" + node.crypto.randomBytes(4).readUInt32LE(0).ToString() + ".md"
            let tempPath = node.path.join (node.os.tmpdir(), fileName)
            // Write the error message to the file, this will allow us to close the text editor 
            // without a confirmation dialog as the file content will not be modified.
            node.fs.writeFileSync(tempPath, msg)
            // Open the file in VSCode
            let newFile = vscode.Uri.parse (tempPath)
            let! document = workspace.openTextDocument newFile
            let! _ =  window.showTextDocument (document, ?options = None)
            // Show the error message in the markdown preview (better display)
            do! commands.executeCommand("markdown.showPreview", Some (box document))
            // Close the text editor (this does not close the markdown preview)
            do! commands.executeCommand("workbench.action.closeActiveEditor", Some (box document))
            // Delete the temporary file as it is not needed anymore.
            node.fs.unlinkSync(U2.Case1 tempPath)
        | Some _ 
        | None ->
            ()
        return failwith "no `dotnet` binary found"
    }
```

MangelMaxime:
Hello @baronfel @Krzysztof-Cieslak,

I rebased this PR to the latest version of main + improved the error display as show in my last message.

For me this PR is ready for review now.

Krzysztof-Cieslak:
Thanks a lot! ❤️

PR #1734, opened 2022-07-18
Add more `when` cases for contributed commands

Krzysztof-Cieslak:
Make sure we don't display commands depending on project files if no project files have been loaded.


baronfel:
awesome man, I had hoped it would be this easy!

baronfel (approved):

PR #1737, opened 2022-07-18
Upgrade fantomas + apply it to all the files

### LINKED ISSUE #1736: Should Fantomas run against all the files when checking the formatting?
Hello,

currently it seems like the build system only execute Fantomas only on the last modified file from HEAD.

https://github.com/ionide/ionide-vscode-fsharp/blob/5e6df4e48668783d79a51e0083e0c8dde6d96089/build/Program.fs#L368-L390

This is kind of a strange behaviour. For example, it means I can commit non valid files (in term of formatting) if I do 2 commits for example and don't run the Format command after each commit. This is something that is happening right now as for example right now all these files are wrongly formatted:

```bash
src/fsharp.fs needs formatting
src/Core/Utils.fs needs formatting
src/Core/FsProjEdit.fs needs formatting
src/Core/Environment.fs needs formatting
src/Core/Project.fs needs formatting
src/Core/LanguageService.fs needs formatting
src/Core/Logging.fs needs formatting
src/Components/Help.fs needs formatting
src/Components/InfoPanel.fs needs formatting
src/Components/QuickInfoProject.fs needs formatting
src/Components/FakeTargetsOutline.fs needs formatting
src/Components/Gitignore.fs needs formatting
src/Components/Debugger.fs needs formatting
src/Components/FSharpLiterate.fs needs formatting
src/Components/TestExplorer.fs needs formatting
src/Components/InlayHints.fs needs formatting
src/Components/ScriptRunner.fs needs formatting
src/Components/QuickInfo.fs needs formatting
src/Components/MSBuild.fs needs formatting
src/Components/CodeLensHelpers.fs needs formatting
src/Components/LineLens.fs needs formatting
src/Components/Diagnostics.fs needs formatting
src/Components/Fsi.fs needs formatting
src/Components/SolutionExplorer.fs needs formatting
src/Components/PipelineHints.fs needs formatting
src/Imports/XPath.fs needs formatting
```

I think it would be more intuitive if we run Fantomas against all the files all the times.

What do you think?

cc @nojaf

MangelMaxime:
Fix #1736 

Changes:

* Upgrade Fantomas to the latest beta version
* Apply Fantomas to all the F# files in `src` and `build` folders.
* Remove unused `open` statements
* Rewrite access to Node API because of a bug in Fantomas (see below)

When executing Fantomas, I had a bunch of error due to `open global.Node` being rewritten to ``open `global`.Node``.

I made the decision to fix this error by changing:

```fs
open global.Node

let myFunc = path.join(..., ...)
```

to 

```fs
module node = Node.Api

let myFunc = node.path.join(..., ...)
```

The `module node = Node.Api` is something that is already used in different places of the code. And so like that all the nodes are now consistent.

The other solutions would be to disable Fantomas for 8-10 files in the project because that bug.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -15,7 +15,7 @@
       ]
     },
     "fantomas": {
-      "version": "5.0.0-alpha-007",
+      "version": "5.0.0-beta-001",
       "commands": [
         "fantomas"
       ]
--- a/build/Program.fs
+++ b/build/Program.fs
@@ -25,9 +25,7 @@ let gitHome = "https://github.com/" + gitOwner
 let gitName = "ionide-vscode-fsharp"
 
 // Read additional information from the release notes document
-let releaseNotesData =
-    File.ReadAllLines "RELEASE_NOTES.md"
-    |> ReleaseNotes.parseAll
+let releaseNotesData = File.ReadAllLines "RELEASE_NOTES.md" |> ReleaseNotes.parseAll
 
 let release = List.head releaseNotesData
 
@@ -99,11 +97,7 @@ module Fable =
 
         let fableProjPath = args.ProjectPath
 
-        let fableDebug =
-            if args.Debug then
-                "--define DEBUG"
-            else
-                ""
+        let fableDebug = if args.Debug then "--define DEBUG" else ""
 
         let fableExperimental =
             if args.Experimental then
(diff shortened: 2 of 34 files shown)


baronfel (approved):
Love it - thank you!

nojaf:
I'm working on that `global` fix btw, but I need https://github.com/dotnet/fsharp/pull/13523 to be merged in the compiler before I can tackle it on Fantomas' side.

MangelMaxime:
Thank you @nojaf that's good news. :)

PR #1751, opened 2022-07-24
Remove old code related to FAKE

MangelMaxime:
Hello @Krzysztof-Cieslak 

When re-basing my PR #1733 on master I found this code that seems to be unused.

I think this is related to FAKE Targets Outline, which was removed in one of your recent commits.

*Sending this change as a separate PR to make it easier to check*


Krzysztof-Cieslak:
Thanks!

PR #1759, opened 2022-08-16
Fix #1073: Prevent hangs on double tick function name

### LINKED ISSUE #1073: Ionide hangs when writing a comment having a certain format.
**Describe the bug**
When writing ~~a comment~~ a text having a format ```//.*``.*\(.*\).*``.+```, Ionide slows down and hangs in the end.

**To Reproduce**
Steps to reproduce the behaviour:
1. Open any F# file.
2. ~~Start commenting.~~
3. Input ``` ``()`` ``` and keep pushing `a`. It will look like ``` //``()``aaaaaaaaaaa...```.
4. VSCode hangs.

**Expected behaviour**
Nothing bad happens.

 - OS: Debian GNU/Linux buster/sid
 - Ionide version: 3.35.0
 - VSCode version: 1.33.1
 - dotnet SDK version: 2.2.203
 - mono / .Net Framework version: Mono 5.20.1.19

**Additional context**
This bug does not happen while using [fsharp-language-server](https://github.com/fsprojects/fsharp-language-server), so I think it is somewhat related to Ionide.

MangelMaxime:
# Eureka ! I found it !

The problem that is causing VSCode to hand on certain word pattern is coming from `wordPattern` configuration in `language-configuration.json`.

https://github.com/ionide/ionide-vscode-fsharp/blob/b2fb2bc2c9cd3ab26f5b1a6a2a361f2edc3188a9/release/language-configuration.json#L50

---

The question is do we need a custom word pattern?

I tested by completely removing that line and it seems like intellisense, moving caret word by word, is still working.

I looked at F# Language Server and it doesn't seems to provide this setting so I think we should be good to remove it and let the default behaviour for now. Especially because having the custom rule `(``((\\w*\\s?)\\s?)*``)` is hurting "that much".

If needed we can always try to re-add it using another regex or explicitly not supporting the double tick function/variable names.


baronfel:
Lovely! Thank you for the detailed investigation and the fix 🎉

PR #1766, opened 2022-08-18
Group tests by namespace across multiple files

kojo12228:
image

From messing around with the Testing panel and looking around the code, it seems Ionide creates a new top level `TestItem` for every file, even if files share a namespace. The screenshot above shows a way to reproduce this behaviour.

I don't exactly understanding the VSCode API, not how testing works in Ionide really, so this PR is kind of a stab at the dark, working on the assumption that common parents would have the same label. Hopefully there is a better approach to this.

Sidenote: Ideally I would test this locally, I tried following the contribution guide, but I keep getting the following exceptions so at the moment I just have my fingers crossed this is the right solution:

```
Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET program, but dotnet---roll-forward does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
```


MangelMaxime:
@kojo12228 What commands are you running for trying to run Ionide locally?

In theory, you want to run `./build.sh -t Build` and then go VSCode tasks and launch the "Watch" one. 
It will say that it fails blabla, you can say "Debug anyway" and it should works.

kojo12228:
@MangelMaxime So the build command `./build Build` in Powershell works:

```
---------------------------------------------------------------------
Build Time Report
---------------------------------------------------------------------
Target          Duration
------          --------
Clean           00:00:00.0323693
DotNetRestore   00:00:02.4574993
YarnInstall     00:00:00.4258284
Format          00:00:02.8018848
RunScript       00:00:20.6346495
Default         00:00:00.0000495
Total:          00:00:26.4865000
Status:         Ok
---------------------------------------------------------------------
```

If I select `Build (watch) and Launch Extension` in the `Run and Debug` panel, I get the `Debug anyway` dialog, which I select. Then when the Extension Development Host VSCode launches, the output shows:

```
[Error - 08:21:52] Client F#: connection to server is erroring. Shutting down server.
[Error - 08:21:52] Stopping server failed
Error: Client is not running and can't be stopped. It's current state is: starting
	at LanguageClient.shutdown (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:17490:19)
	at LanguageClient.stop (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:17461:21)
	at LanguageClient.stop (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:26486:22)
	at LanguageClient.handleConnectionError (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:17697:18)
	at errorHandler (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:17640:18)
	at d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:18010:36
	at CallbackList.invoke (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:14569:39)
	at Emitter.fire (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:14631:36)
	at writeErrorHandler (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:13603:22)
	at CallbackList.invoke (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:14569:39)
	at Emitter.fire (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:14631:36)
	at StreamMessageWriter.fireError (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:15509:27)
	at Socket.<anonymous> (d:\Code\Forked Repos\ionide-vscode-fsharp\release\fsharp.js:15546:47)
	at Socket.emit (node:events:390:28)
	at emitErrorNT (node:internal/streams/destroy:157:8)
	at emitErrorCloseNT (node:internal/streams/destroy:122:3)
	at processTicksAndRejections (node:internal/process/task_queues:83:21)
Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET program, but dotnet---roll-forward does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
```

Eventually I get

```
[Error - 08:21:53] The F# server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.
```

Same thing happens for all the other "* and Launch Extension" launch configurations.

MangelMaxime:
@kojo12228 You need to run the command `./build.sh -t Build` (there `-t` is important).

When you do `./build.sh Build`, it is not running `Build` target but `Default`. You can confirm it by looking at the list of target that has been executed in your first snippet.

Running `./build.sh -t Build`, will build Ionide and most importantly copy the FSAC (LSP server) server executable for Ionide to run it. Without those executable you get the error about "crashed 5 times ..."

kojo12228:
Worked perfectly @MangelMaxime thanks. And the feature seems to work as expected, although checking equality on the label feels kind of hacky? Any idea on a better solution? I'll look around the API see if there is something that makes more sense, without a big change in FSAC (although maybe that's the right thing to do).

baronfel:
@kojo12228 This seems like a really nice change - thanks for looking into it! I agree that the grouping you've proposed makes a ton of sense. We can go with this, and if grouping on the label isn't the right answer long term we can revisit it :)

PR #1780, opened 2022-09-10
Fix CONTRIBUTING links

marcelocra:
Currently some of them are not pointing to the correct place.

Notice that in the beginning of the file there is the following, which looks like links that were not applied because of missing md related parts:

image


baronfel:
Thanks for this!

PR #1781, opened 2022-09-13
Added settings for AdaptiveLspServer in FSAC

TheAngryByrd:
Flag to allow https://github.com/fsharp/FsAutoComplete/pull/1007 to be used in ionide


baronfel (changes requested):
This LGTM, but let's wait to merge until https://github.com/fsharp/FsAutoComplete/pull/1007 merges and releases, then we can update FSAC itself in this PR as well. That way there's never a point in time where a use can select Ionide options that will crash the FSAC instance.

baronfel:
Very happy to have this in.

baronfel (approved):

PR #1782, opened 2022-09-23
Get changes in place to use multitargeted FSAC and pivot accordingly.

baronfel:
We'll need to bump FSAC again because the 57.0 release didn't have the net7.0 TFM bits, but that change is already done - just needs a 0.57.1 bump.

### DIFF
--- a/build/Program.fs
+++ b/build/Program.fs
@@ -311,13 +311,17 @@ let initTargets () =
 
 
     Target.create "CopyFSACNetcore" (fun _ ->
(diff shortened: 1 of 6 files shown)


baronfel:
Ok, FSAC bumped again and deployments fixed. I've tested the 6/7 SDK selection and that's working great.

baronfel:
Going to go ahead and merge this - if we need to take up another FSAC bump that'll be a normal bump PR from this point forward.

PR #1792, opened 2022-10-16
Quickfix for overloading server with FSharpSignatureData requests

TheAngryByrd:
👋 I noticed there's been an issue with larger files getting bombarded with `FSharpSignatureData`.  Since this was firing a lot of promises at the same time, this was causing VSCode to slow down to a crawl. 

While it would be better to only send up ranges you're currently viewing, this is a quick fix to try to stop the bleeding.

### DIFF
--- a/src/Components/LineLens.fs
+++ b/src/Components/LineLens.fs
@@ -195,8 +195,10 @@ module DecorationUpdate =
         textLine.range
 
     let private getSignature (uri: Uri) (range: DTO.Range) =
-        promise {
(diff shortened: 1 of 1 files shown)


baronfel:
This is a good quick fix. Longer term I'd like to make the LSP endpoint for signature data take a range to compute the lenses for (as you know, I'm just repeating it here for the record).

TheAngryByrd (commented):
~~ignore~~

PR #1826, opened 2023-02-19
Adds FSharp.fsac.converveMemory flag

TheAngryByrd:
This currently sets [DOTNET_GCConserveMemory ](https://learn.microsoft.com/en-us/dotnet/core/runtime-config/garbage-collector#conserve-memory) which should help with Large Object Heap allocations in FCS. Definitely another bandaid in a long line of bandaids but it does seem to help with larger projects.

image

image

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -516,7 +516,7 @@
         },
         "FSharp.fsac.conserveMemory" : {
           "default" : false,
-          "description": "This will set DOTNET_GCConserveMemory. Requires restart.",
+          "description": "Configures FsAutoComplete with settings intended to reduce memory consumption. Requires restart.",
           "type" : "boolean"
         },
         "FSharp.fsac.dotnetArgs": {
--- a/release/package.json
+++ b/release/package.json
@@ -514,6 +514,11 @@
           "description": "Appends the \u0027--attachdebugger\u0027 argument to fsac, this will allow you to attach a debugger.",
           "type": "boolean"
         },
+        "FSharp.fsac.conserveMemory" : {
+          "default" : false,
+          "description": "This will set DOTNET_GCConserveMemory. Requires restart.",
+          "type" : "boolean"
+        },
         "FSharp.fsac.dotnetArgs": {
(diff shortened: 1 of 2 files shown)


release/package.json
```diff
@@ -514,6 +514,11 @@
           "description": "Appends the \u0027--attachdebugger\u0027 argument to fsac, this will allow you to attach a debugger.",
           "type": "boolean"
         },
+        "FSharp.fsac.conserveMemory" : {
+          "default" : false,
+          "description": "This will set DOTNET_GCConserveMemory. Requires restart.",
```
baronfel, line 519:
```suggestion
          "description": "Configures FsAutoComplete with settings intended to reduce memory consumption. Requires restart.",
```

src/Core/LanguageService.fs
```diff
@@ -678,6 +678,8 @@ Consider:
             let enableProjectGraph =
                 "FSharp.enableMSBuildProjectGraph" |> Configuration.get false
 
+            let convserveMemory = "FSharp.fsac.conserveMemory" |> Configuration.get false
```
baronfel, line 681:
```suggestion
            let conserveMemory = "FSharp.fsac.conserveMemory" |> Configuration.get false
```

src/Core/LanguageService.fs
```diff
@@ -766,6 +768,11 @@ Consider:
                 promise {
                     let! (fsacDotnetArgs, fsacEnvVars, fsacPath) = discoverDotnetArgs ()
 
+                    let fsacEnvVars =
+                        [ yield! fsacEnvVars
+                          if convserveMemory then
```
baronfel, line 773:
```suggestion
                          if conserveMemory then
```

baronfel (approved):
Minor rewording nitpicks, but otherwise LGTM!

baronfel:
Can you add this to the configuration docs at ionide.io as well? That would be the appropriate place for a list of what knobs get tuned, like linking to this flag.

PR #1831, opened 2023-02-25
OpenTelemetry settings

TheAngryByrd:
Corresponding changes related to https://github.com/fsharp/FsAutoComplete/pull/1060, only merge after FSAC changes have been merged.


release/package.json
```diff
@@ -665,6 +665,25 @@
...
+        "FSharp.notifications.traceNamespaces": {
+          "default": [
+            "BoundModel.TypeCheck",
+            "BackgroundCompiler."
+          ],
+          "description": "The set of System.Diagnostics.Activity names to watch.",
+          "type" : "array",
+          "required": ["FSharp.notifications.trace"]
+        },
+        "FSharp.openTelemetry.enabled": {
```
baronfel, line 682: This setting needs to be true in order for `notifications.trace` to have any effect, right? If so, I'd add that to the description, and then add that that check to the LanguageService CLI arg condition too.

TheAngryByrd: No, the notifications uses an `ActivityListener` which is independent of export OTel traces.

baronfel: oh, perfect :+1:

baronfel:
Tested locally, works great :thumbsup:

PR #1833, opened 2023-02-26
Update FSAC to get F# 7 and Opentelemetry fixes

baronfel:


.config/dotnet-tools.json
```diff
@@ -9,13 +9,13 @@
       ]
     },
     "fable": {
-      "version": "3.7.11",
+      "version": "4.0.0-theta-018",
```
baronfel, line 12: since I bumped the global.json to .NET 7, I needed a Fable that supported F# 7.

release/package.json
```diff
@@ -615,6 +612,16 @@
...
+        "FSharp.inlineValues.enabled": {
+          "default": false,
+          "description": "Enables rendering all kinds of hints inline with your code. Currently supports pipelineHints, which are like LineLenses that appear along each step of a chain of piped expressions",
+          "type": "boolean"
+        },
+        "FSharp.inlineValues.prefix": {
+          "default": "  // ",
+          "description": "The prefix used when rendering inline values.",
+          "type": "string"
+        },
```
baronfel, line 624: Here are the new inlineValues settings to match the config in FSAC - these default to 'off' because the FSAC mechanism isn't fully working for some reason. We maybe should comment them out/remove them entirely for this release until the inlineValues mechanism is proven to be working, but most folks don't mess with settings so it should be safe.

baronfel:
Once this merges we can rebase @TheAngryByrd's otel settings PR and we should be good to go for a new release.

PR #1836, opened 2023-03-05
Test Explorer: support test cases

### LINKED ISSUE #1756: Test Explorer doesn't support tests with `TestCaseAttribute`
### Describe the bug

Test explorer doesn't work for NUnit tests that use `TestCaseAttribute`.

- Test builds and runs (.NET process running test ends)
- VSCode spinning icon does not go away until test is cancelled

### Steps to reproduce

1. Create a new NUnit test project (`dotnet new nunit -lang F#`)
2. Open directory in Visual Studio Code
3. In `UnitTest1`, replace `[<Test>]` with `[<TestCase("test")]`
4. Select `Run test` in Test Explorer, or ▶️ in gutter

### Expected behaviour

VSCode should correctly show the status of the test (whether succeeded or failed). Similarly, the Test Explorer should break down those test cases, rather than just showing one test.

### Screenshots

test_explorer_example

## Machine info
 - OS: Windows
 - .NET SDK version: 6.0.302
 - Ionide version: 7.0.0

---

Frustratingly, I couldn't find anything in the logs that was logged as Debug in `TestExplorer.fs`, otherwise I might have found it easier to find out what parsing of tests/test results is failing. I would like to take a stab at trying to fix this, just depends on time (and whether I get distracted by other projects).

kojo12228:
Add support in VSCode Test Explorer for showing multiple tests that belong to one method.

Implementation details:

- Assume project test items from FSAutoComplete may actually refer to multiple tests
- Sort project tests by longest name and find test results that start with that test name.
  - Sorting ensures that a test method that is a substring of another doesn't get associated with the longer method name's tests.
- If more than one test is found for that project test, nest all results under that project test

Fixes #1756

test_explorer_test_cases_example

### DIFF
--- a/src/Components/TestExplorer.fs
+++ b/src/Components/TestExplorer.fs
@@ -16,7 +16,7 @@ let private lastOutput = Collections.Generic.Dictionary<string, string>()
 let private outputChannel = window.createOutputChannel "F# - Test Adapter"
 
 let private logger =
-    ConsoleAndOutputChannelLogger(Some "TestExplorer", Level.DEBUG, None, Some Level.DEBUG)
+    ConsoleAndOutputChannelLogger(Some "TestExplorer", Level.DEBUG, Some outputChannel, Some Level.DEBUG)
 
 type TestItemCollection with
 
@@ -67,7 +67,7 @@ type ProjectWithTestResults =
       Tests: TestResult array }
 
 module DotnetTest =
-    let runProject (projectWithTests: ProjectWithTests) : JS.Promise<ProjectWithTestResults> =
+    let runProject (tc: TestController) (projectWithTests: ProjectWithTests) : JS.Promise<ProjectWithTestResults> =
         logger.Debug("Nunit project", projectWithTests)
 
         promise {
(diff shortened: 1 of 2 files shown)


src/Components/TestExplorer.fs
```diff
@@ -131,59 +131,91 @@ module DotnetTest =
...
+
+                        { Test = t.Test
+                          FullTestName = fullName
+                          Outcome = !!outcome
+                          ErrorMessage = errorInfoMessage
+                          ErrorStackTrace = errorStackTrace
+                          Expected = expected
+                          Actual = actual
+                          Timing = timing }))
+                |> Array.concat
```
baronfel, line 218: nit: normally a map + concat can be rewritten to use `collect` instead - can you check into that?

src/Components/TestExplorer.fs
```diff
@@ -131,59 +131,91 @@ module DotnetTest =
...
-                    let errorInfoMessage =
-                        xpathSelector.TrySelectString
-                            $"/t:TestRun/t:Results/t:UnitTestResult[@executionId='{executionId}']/t:Output/t:ErrorInfo/t:Message"
+            let unmappedTestNames =
+                unmappedTests |> Array.map (fun (fullName, _, _) -> fullName)
 
-                    let errorStackTrace =
-                        xpathSelector.TrySelectString
-                            $"/t:TestRun/t:Results/t:UnitTestResult[@executionId='{executionId}']/t:Output/t:ErrorInfo/t:StackTrace"
+            logger.Debug("Unmapped tests", unmappedTestNames)
```
baronfel, line 158: it looks like we do this second array walk only to emit the names of the unmapped tests - do you see this happening anywhere, or was this a debugging aid while you were developing? if the latter, can you delete it so we do less work?

baronfel (approved):
This looks great overall - a wonderful investigation and fix. I left a couple comments on the implementation, can you take a look and tell me what you think?

src/Components/TestExplorer.fs
kojo12228: Not that I saw this happening, but I'm pretty sure this implementation is a plaster over the issue rather than the right fix. The intention was to add it in to make it easier to review which tests don't get correctly mapped in the implementation, but given it's a few lines and walks over the structure, I'll remove it.

kojo12228: https://github.com/ionide/ionide-vscode-fsharp/issues/1756#issuecomment-1445485152 This example I don't think is fixed by this implementation, but maybe that could be addressed in a subsequent PR.

baronfel: I agree - happy to take incremental improvements to this as they come!

baronfel:
I hope to merge this this week - there are a couple changes to FSAC I want to grab as well.

PR #1837, opened 2023-03-05
update callsites for FSAC being option-aware now

baronfel:
These commits are reactions to https://github.com/fsharp/FsAutoComplete/pull/1069's changes in the response shapes of the custom PlainNotification endpoints - it turns out that many of the endpoints were already null-aware, but this made it more explicit.


baronfel:
This works well with existing FSAC as well, so I'll merge.

PR #1841, opened 2023-03-09
Use base64 encoded name in test URIs

kojo12228:
Currently the test URI is based upon the ID from the TestAdapterEntry from FsAutoComplete. However, this ID is based on the order of tests, and not based on the content. Thus, VSCode can incorrectly assign status to an unrelated test when the ordering of tests is changed, see below:

test_explorer_uri_master

This change uses the name of the test entry (base64 encoded) instead, which means that provided the test name is fixed, VSCode will continue to assign the correct status to that test:

test_explorer_uri_new

I should note, there is a benefit of the existing ID based approach in that if a test is renamed, but is otherwise unchanged, VSCode will continue to assign the same status. I would suggest, however, that it make more sense to change the test URI in that case.

### DIFF
--- a/src/Components/TestExplorer.fs
+++ b/src/Components/TestExplorer.fs
@@ -1,6 +1,7 @@
 module Ionide.VSCode.FSharp.TestExplorer
 
 open System
+open System.Text
 open Fable.Core
 open Fable.Import.VSCode
 open Fable.Import.VSCode.Vscode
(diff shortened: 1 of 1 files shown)


baronfel:
how does this sequence with #1844 - is there a particular order required for merging them?

kojo12228:
This PR is unrelated to #1844 and I would argue this should be merged first, since I'm more confident on this change. I can then I can deal with the conflict after.

baronfel:
Perfect, let's go with that then.

PR #1844, opened 2023-03-11
Handle tests within nested modules

### LINKED ISSUE #1756: Test Explorer doesn't support tests with `TestCaseAttribute`

kojo12228:
Fixes #1756 

Using module type information from https://github.com/fsharp/FsAutoComplete/pull/1071, correctly handle the full name of tests where there is a module within a module, a type within a module or a module which shares the same name as a sibling type.

test_explorer_nested_modules

Unfortunately, `TestAdapterEntry` which holds the information about module type is not accessible at the point where the full name is constructed, and the VSCode type `TestItem` has no field where this information can be added (AFAIK). To workaround that, this PR uses an otherwise undesirable `ResizeArray` acting as a map of sorts to hold mutable state about module types and this is passed to various functions. Ideally, something like a Set would be used but I will double check if Fable supports `System.Collections.Generic.HashSet` (otherwise it might be worth doing some housekeeping to ensure this collection doesn't become a memory leak).

### DIFF
--- a/paket.lock
+++ b/paket.lock
@@ -242,4 +242,4 @@ STORAGE: PACKAGES
 RESTRICTION: == netstandard2.0
 NUGET
   remote: https://api.nuget.org/v3/index.json
-    fsautocomplete (0.59.1)
+    fsautocomplete (0.59.2)
--- a/src/Components/Help.fs
+++ b/src/Components/Help.fs
@@ -14,6 +14,7 @@ module Help =
 
         promise {
             let! res = LanguageService.f1Help (doc.uri) (int pos.line) (int pos.character)
+
             match res with
             | None -> return ()
             | Some res ->
--- a/src/Components/InfoPanel.fs
+++ b/src/Components/InfoPanel.fs
@@ -196,9 +196,7 @@ module InfoPanel =
                     let doc = textEditor.document
                     let pos = selections.[0].active
                     let! res = LanguageService.documentation doc.uri (int pos.line) (int pos.character)
-                    res
-                    |> Option.bind mapContent
-                    |> Option.iter setContent
+                    res |> Option.bind mapContent |> Option.iter setContent
                 else
                     return ()
             }
@@ -209,9 +207,7 @@ module InfoPanel =
                 if isFsharpTextEditor textEditor && panel.IsSome then
                     let doc = textEditor.document
                     let! res = LanguageService.documentation doc.uri (int pos.line) (int pos.character)
-                    res
-                    |> Option.bind mapContent
-                    |> Option.iter setContent
+                    res |> Option.bind mapContent |> Option.iter setContent
                 else
                     return ()
             }
@@ -220,9 +216,7 @@ module InfoPanel =
         let updateOnLink xmlSig assemblyName =
             promise {
                 let! res = LanguageService.documentationForSymbol xmlSig assemblyName
-                res
-                |> Option.bind mapContent
-                |> Option.iter setContent
+                res |> Option.bind mapContent |> Option.iter setContent
             }
 
     let mutable private timer = None
--- a/src/Components/LineLens.fs
+++ b/src/Components/LineLens.fs
@@ -196,8 +196,7 @@ module DecorationUpdate =
 
     let private getSignature (uri: Uri) (range: DTO.Range) =
         promise {
-            let! signaturesResult =
-                LanguageService.signatureData uri range.StartLine (range.StartColumn - 1)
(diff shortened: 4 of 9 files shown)


baronfel:
Great job! I'll wait to merge this until the matching FSAC PR is merged and released, ok?

baronfel:
@kojo12228 can you resolve the conflicts with the other PR when you have a moment? I'm working on an FSAC release so when that's completed I'm ready to merge and release this fix as well.

kojo12228:
@baronfel sorry for the delay, I merged them locally but was getting issues with test explorer that I don't get on `main`. I'll push the merge change for now

kojo12228:
## Odd behaviour since merging

For further details, tests seems to be correctly being discovered with [eea849e](https://github.com/ionide/ionide-vscode-fsharp/pull/1844/commits/eea849eaaaf3ab2dab71c43e4a4634f7aa91c7c0):

test_explorer_nested_modules_eea849e

However something is happening on [182b6ff](https://github.com/ionide/ionide-vscode-fsharp/pull/1844/commits/182b6ff285573d694b67e8d8a0bc94c1ec014d88) that I can't explain, `UnitTest1.fs` and `UnitTest3.fs` tests don't get correctly discovered. They're using the same version of FSAC:

test_explorer_nested_modules_182b6ff

### The diff between the commits (ignoring commits) is the following

Starting line 194 in this PR:
```diff
                        if Seq.length testCases > 1 then
                            let ti =
-                                t.Test.children.get (t.Test.uri.Value.ToString() + " -- " + testName)
+                                t.Test.children.get (
+                                    t.Test.uri.Value.ToString()
+                                    + " -- "
+                                    + Convert.ToBase64String(Encoding.UTF8.GetBytes(testName))
+                                )
                                |> Option.defaultWith (fun () ->
                                    tc.createTestItem (
```

Starting line 224 in this PR:
```diff
let rec mapTest
    (tc: TestController)
    (uri: Uri)
    (moduleTypes: Collections.Generic.Dictionary<string, string>)
    (t: TestAdapterEntry)
    : TestItem =
-    let ti = tc.createTestItem (uri.ToString() + " -- " + string t.id, t.name, uri)
+    let ti =
+        tc.createTestItem (
+            uri.ToString() + " -- " + Convert.ToBase64String(Encoding.UTF8.GetBytes(t.name)),
+            t.name,
+            uri
+        )
```

Starting line 271 in this PR:
```diff
-    logger.Debug("Module types", moduleTypes)

    let rec getFullName (ti: TestItem) =
```

kojo12228:
Also `main` works without issue as well, so it's something with interleaving of the two that's the issue. I find that weird, especially because I originally wrote the first commit on top of #1841

kojo12228:
Figured it out, #1841 unintentionally made a change such that `TestItem.id` was no longer unique across an entire source file (due to `t.id` being unique amongst the file), as it was based on the name of the method/module/namespace and nothing else. The following would silently fail:

```
module Outer =
    [<TestFixture>]
    module Inner =
        [<Test>]
        let Test1 () = // First Test1
            Assert.Pass ()

        module Innermost =
            [<Test>]
            let Test1 () = // Second Test1, Dictionary.Add would silently throw
                Assert.Pass()
```

Fixed this by passing a `parentNameId` string down the tree and including it in the `TestItem.id`.

I think we're good to go @baronfel , apologies for the delay.

baronfel:
No worries at all, thank you for taking the time to dig into it :) I'll try to get out a new release containing all of this later today!

kojo12228:
@baronfel works as expected with FsAutoComplete 0.59.2 (not sure if you want the commit in this branch though)

PR #1845, opened 2023-03-11
Support 'Add file above/below' from folders

### LINKED ISSUE #1825: Add file does not work for files in folders
### Describe the bug

In the F# Solution Explorer, when right clicking on a File inside a folder, I can select "Add file above" or "Add file below", but nothing happens.  I also cannot get a context menu at all for folders.

#### Case 1

* Navigate to the F# Solution Explorer
* Locate a file nested inside a folder

Screenshot_20230217_123402

* Right click any file inside that folder to show the context menu

Screenshot_20230217_123526

* Click "Add file below" or "Add file above"
* Nothing happens

#### Case 2

* Navigate to the F# Solution Explorer
* Locate a file nested inside a folder
* Right click any folder to show the context menu
* Nothing happens

### Expected behaviour

The add file dialog appears

## Machine info
 - OS: Windows
 - .NET SDK version: 7.0.102
 - Ionide version: 7.4.2

MangelMaxime:
Fix #1825

### DIFF
--- a/src/Components/SolutionExplorer.fs
+++ b/src/Components/SolutionExplorer.fs
@@ -662,13 +662,51 @@ module SolutionExplorer =
 
         window.showInputBox opts
 
+    /// <summary>
+    /// Try to find the parent project of a given model.
+    ///
+    /// Important: If the given model is a project, it will return the model itself.
+    /// </summary>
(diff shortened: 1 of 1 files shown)


baronfel:
Great find :+1:

PR #1846, opened 2023-03-11
Provide default fileNesting patterns

### LINKED ISSUE #1724: Provide default settings for VSCode Explorer File Nesting feature
VSCode 1.67 provides a new File Nesting feature: [docs](https://code.visualstudio.com/updates/v1_67#_explorer-file-nesting)

We should provide default `explorer.fileNesting.patterns` for F# files, so that when users turn the feature on via `explorer.fileNesting.enabled` F# just works.

Proposed mappings:

```json
{
  "contributes": {
    "configurationDefaults": {
      "[fsharp]": {
        "explorer.fileNesting.patterns": {
          "*.fs": "${basename}.fsi, ${basename}.fs.js, ${basename}.fs.js.map",
          "*.fsl": "${basename}.fsi, ${basename}.fs",
          "*.fsy": "${basename}.fsi, ${basename}.fs"
        }
      }
    }
  }
}
```

MangelMaxime:
Fix #1724

Adding the configuration like done in this PR seems to be enough to have nesting pattern added for F#.

This configuration is merged with the default configuration from VSCode:

image

The picture above is the result of doing `Fable.Core.JS.console.log (Configuration.getUnsafe "explorer.fileNesting.patterns")` when loading Ionide.

However, the drawback I see is that if the user don't want the nested feature for F# files and he try to set its configuration then the default value completed doesn't contain the default configuration from VSCode but only the ones from Ionide.

https://user-images.githubusercontent.com/4760796/224512424-f4d35e79-99eb-406f-892a-ba2e894b604f.mov

Personally, I think this is problem in VSCode and not in Ionide.

If we want, to preserve the defaults configuration then, we would need to do something similar to **Suggest gitignore**:

- Ask the user if he wants F# nesting patterns. 
- If yes
    - Retrieve the current nesting patterns settings
    - Merge with F# nesting patterns and save in the workspace settings
- If no, discard or save user choice to not suggest again

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -38,6 +38,13 @@
         "language": "fsharp"
       }
     ],
+    "configurationDefaults": {
+      "explorer.fileNesting.patterns": {
+        "*.fs": "${basename}.fsi, ${basename}.fs.js, ${basename}.fs.js.map",
+        "*.fsl": "${basename}.fsi, ${basename}.fs",
+        "*.fsy": "${basename}.fsi, ${basename}.fs"
+      }
+    },
     "colors": [
       {
         "defaults": {


baronfel (approved):
Thanks for doing this :)

MangelMaxime:
Always happy to help :)

Doing a tour of the issues I can fix in Ionide.

PR #1848, opened 2023-03-12
Add "Rename file" to the solution explorer

MangelMaxime:
This PR needs https://github.com/fsharp/FsAutoComplete/pull/1075

This PR adds:

- Rename file to the contextual menu of the solution explorer
- Create a new group in contextual menu of a file which contains "actions on the file"
    - Rename file
    - Remove file
    
    I did that because I found that otherwise it was a bit too clamp.
- Make the file name detection mirror the logic of the commands

    If the user provide a file name without a F# extension, the commands add the `.fs` extension automatically. The validation of the user input now follow the same logic.

    Is it the correct way of doing it? Fsproj are not limited to `.fs` / F# files, so perhaps we should respect what the user provided.

https://user-images.githubusercontent.com/4760796/224574795-d553e4ed-ef24-4cc7-aed1-d8fd1064846d.mov

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -208,6 +208,10 @@
         "title": "Move file down"
       },
       {
+        "command": "fsharp.explorer.renameFile",
+        "title": "Rename file"
+      },
+      {
         "command": "fsharp.explorer.removeFile",
         "title": "Remove file"
       },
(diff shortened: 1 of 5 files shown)


baronfel:
Will merge when a new FSAC is pushed (hopefully today!)

baronfel (approved):

release/package.json
```diff
@@ -921,6 +925,10 @@
           "command": "fsharp.explorer.moveDown",
           "when": "false"
         },
+        {
+          "command": "fsharp.explorer.renameFile",
+          "when": "false"
```
baronfel, line 930: should this have a different condition? or is it only intended to be triggered from the explorer?

PR #1873, opened 2023-05-16
Don't type test against interfaces

alfonsogarciacaro:
See this comment: https://github.com/fable-compiler/Fable/issues/3443#issuecomment-1549617520

Fable converts pattern matching with erased unions (like `U2`) to type testing, but this doesn't work all the time. For example, it's not possible to type test against interfaces because they are not actually declared in the final JS code. Ideally we should use `Choice` instead of an erased union in those cases but here the values come from VS Code so I've replaced the pattern matching with other solutions to prevent the compilation warnings and ensure expected behavior.

### DIFF
--- a/src/Components/Fsi.fs
+++ b/src/Components/Fsi.fs
@@ -420,8 +420,7 @@ module Fsi =
             let! profile =
                 match provider.provideTerminalProfile (ctok) with
                 | None -> promise.Return None
-                | Some(U2.Case1 options) -> promise.Return(Some options)
-                | Some(U2.Case2 work) -> Promise.ofThenable work
+                | Some work -> work |> Promise.ofMaybeThenable Some
 
             let profile =
                 match profile with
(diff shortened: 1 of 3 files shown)


src/Components/Fsi.fs
```diff
@@ -431,10 +430,7 @@ module Fsi =
 
                     failwith "unable to spawn FSI"
 
-            let terminal =
-                match profile.options with
-                | U2.Case1 opts -> window.createTerminal opts
-                | U2.Case2 opts -> window.createTerminal opts
```
baronfel, line 437: Ok, so in the general case when we're using erased unions to model VSCode APIs, when 'matching' on those APIs we can't directly match on the case and should instead use methods like probing for specific properties to determine which 'case' we're in?

MangelMaxime: You are right.

You can use type test for native types like `string`.

And for others types, you need to do it as if you were writing JavaScript and so by checking something unique like the constructor instance, a property presence, etc.

You can make it looks nice in F# by using custom active pattern.

alfonsogarciacaro: It's... complicated. As @MangelMaxime says, if you have an erased union with primitives like `U2<string, number>` or actual types `U2<ClassA, ClassB>` runtime type testing will work, but it's not easy to test an structural interface at runtime so for now Fable just emits false in those cases.

We could try to emulate interface type testing by checking the property descriptors of the object, though it still won't work in 100% of cases.

baronfel:
Oh we should have merged this a while back. Thanks again @alfonsogarciacaro!

PR #1883, opened 2023-06-06
Update release package.json

Darkle:
### WHAT
Update the text color for warnings shown in the status bar when using a light theme. 

### WHY
Previously it used #ffcc00, which looks fine with a dark theme, but is very difficult to read when you have a light theme. This pr changes the warning text color for a light theme to #7c6400.

### HOW
Previously it looked like this when using a light theme:

With this change it now looks like this when using a light theme:

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -70,7 +70,7 @@
         "defaults": {
           "dark": "#FFCC00",
           "highContrast": "#FFFF00",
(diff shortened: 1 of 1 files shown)


Darkle:
Actually this change only makes it look ok when the status bar is light. Some light themes have a dark status bar, so this change wont work.

PR #1887, opened 2023-06-18
Adds exclusion config for analyzers

TheAngryByrd:
### <samp>🤖 Generated by Copilot at 5e707e0</samp>

Add configuration options to exclude files from analyzers. Update `./release/package.json` formatting.

### <samp>🤖 Generated by Copilot at 5e707e0</samp>

> _If you use the FSharp extension_
> _You might want some more customization_
> _To exclude some files_
> _From the analyzers' eyes_
> _You can use regex for exclusion_

:sparkles::wrench::page_facing_up:

### WHY

[Corresponding PR on FSAC](https://github.com/fsharp/FsAutoComplete/pull/1120)

### <samp>🤖 Generated by Copilot at 5e707e0</samp>

*  Add configuration options to exclude files from analyzers ([link](https://github.com/ionide/ionide-vscode-fsharp/pull/1887/files?diff=unified&w=0#diff-7902ed52e6f39559068a38e3c223d7f7d7d0b0977aee33d659db648092a62b15R761-R774), [link](https://github.com/ionide/ionide-vscode-fsharp/pull/1887/files?diff=unified&w=0#diff-7902ed52e6f39559068a38e3c223d7f7d7d0b0977aee33d659db648092a62b15R816-R829), [link](https://github.com/ionide/ionide-vscode-fsharp/pull/1887/files?diff=unified&w=0#diff-7902ed52e6f39559068a38e3c223d7f7d7d0b0977aee33d659db648092a62b15R840-R854))
* Add newline at the end of `package.json` ([link](https://github.com/ionide/ionide-vscode-fsharp/pull/1887/files?diff=unified&w=0#diff-7902ed52e6f39559068a38e3c223d7f7d7d0b0977aee33d659db648092a62b15L1662-R1705))

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -758,6 +758,20 @@
           "description": "Enables detection of cases when names of functions and values can be simplified",
           "type": "boolean"
         },
+        "FSharp.simplifyNameAnalyzerExclusions": {
+          "default": [
+            ".*\\.g\\.fs",
+            ".*\\.cg\\.fs"
+          ],
+          "description": "A set of regex patterns to exclude from the simplify name analyzer",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "required": [
+            "FSharp.simplifyNameAnalyzer"
+          ]
+        },
         "FSharp.smartIndent": {
           "default": false,
           "description": "Enables smart indent feature",
@@ -799,6 +813,20 @@
           "description": "Enables detection of unused declarations",
           "type": "boolean"
         },
+        "FSharp.unusedDeclarationsAnalyzerExclusions": {
+          "default": [
+            ".*\\.g\\.fs",
+            ".*\\.cg\\.fs"
+          ],
+          "description": "A set of regex patterns to exclude from the unused declarations analyzer",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "required": [
+            "FSharp.unusedDeclarationsAnalyzer"
+          ]
+        },
         "FSharp.addPrivateAccessModifier": {
           "default": false,
           "description": "Enables a codefix that adds a private access modifier",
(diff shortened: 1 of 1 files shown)


release/package.json
```diff
@@ -758,6 +758,20 @@
           "description": "Enables detection of cases when names of functions and values can be simplified",
           "type": "boolean"
         },
+        "FSharp.simplifyNameAnalyzerExclusions": {
+          "default": [
+            ".*\\.g\\.fs",
+            ".*\\.cg\\.fs"
```
TheAngryByrd, line 764: Since the issue was based on code generated files, this is the conventions I've seen in the wild for good defaults.

baronfel: thought for this - we should probably exclude anything in the intermediate output directory for the project. How are these regexes uses? Are they evaluated per-project?

baronfel (approved):
LGTM, just one question about intermediate outputs

release/package.json
TheAngryByrd: The analyzers are run after each typecheck of a file and this is added to the conditional whether the analyzer is active or doesn’t match one of these reflexes.

TheAngryByrd: So if we wanted to do obj/*.fs we could add it easily.

baronfel: I think intermediate dir would have to be more 'dynamic' or FSAC-included, because it's an MSBuild property that can and does change. .NET 8 adds a new optional directory layout that drastically changes it, for example.

TheAngryByrd: Other option is to just include things we _know_ shouldn't be analyzed like AssemblyInfo.fs.  Otherwise I think this may need more changes to FSAC to get a more dynamic Intermediate folder knowledge.

PR #1903, opened 2023-07-30
Crazy idea to help F# folks getting started - break C# dependency

baronfel:
### WHAT
copilot:summary

copilot:poem

copilot:emoji

### WHY

Ok, so we have a nonzero amount of folks that just want to get going with F#, and right now our hard dependency on the C# extension can introduce a source of instability. What if we could detect the presence of the C# extension at run time and light up debug/launch tasks support if it is found, while notifying the user that those features will not be available until they install that extension?

Basic plan of attack:
* make a new Context value for debugger support
* set that value by probing for the C# extension
* change debugger/msbuild activation to only register commands if C# is present
* use the context value to act as part of the `when` condition for lots of the commands we declare in package.json

TODO:
* [ ] manual testing
* [ ] figure out how the user dialog when no C# extension is available should look
* [ ] figure out how to detect the extensions' runtime context and swap between the openVSIX gallery C# extension and the VSCode marketplace C# extension
* [ ] respond to new extension installations over the lifetime of the Ionide extension

### HOW
copilot:walkthrough

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -155,43 +155,53 @@
       },
       {
         "command": "MSBuild.buildCurrent",
-        "title": "MSBuild: Build current project"
+        "title": "MSBuild: Build current project",
+        "when": "fsharp.debuggerAvailable"
       },
       {
         "command": "MSBuild.buildSelected",
-        "title": "MSBuild: Build project"
+        "title": "MSBuild: Build project",
+        "when": "fsharp.debuggerAvailable"
       },
       {
         "command": "MSBuild.rebuildCurrent",
-        "title": "MSBuild: Rebuild current project"
+        "title": "MSBuild: Rebuild current project",
+        "when": "fsharp.debuggerAvailable"
       },
       {
         "command": "MSBuild.rebuildSelected",
-        "title": "MSBuild: Rebuild project"
+        "title": "MSBuild: Rebuild project",
+        "when": "fsharp.debuggerAvailable"
       },
       {
         "command": "MSBuild.cleanCurrent",
-        "title": "MSBuild: Clean current project"
+        "title": "MSBuild: Clean current project",
+        "when": "fsharp.debuggerAvailable"
       },
       {
         "command": "MSBuild.cleanSelected",
-        "title": "MSBuild: Clean project"
+        "title": "MSBuild: Clean project",
+        "when": "fsharp.debuggerAvailable"
       },
       {
         "command": "MSBuild.restoreSelected",
-        "title": "MSBuild: Restore project"
+        "title": "MSBuild: Restore project",
+        "when": "fsharp.debuggerAvailable"
       },
       {
         "command": "MSBuild.buildCurrentSolution",
-        "title": "MSBuild: Build current solution"
+        "title": "MSBuild: Build current solution",
+        "when": "fsharp.debuggerAvailable"
       },
       {
         "command": "MSBuild.rebuildCurrentSolution",
-        "title": "MSBuild: Rebuild current solution"
+        "title": "MSBuild: Rebuild current solution",
+        "when": "fsharp.debuggerAvailable"
       },
       {
         "command": "MSBuild.cleanCurrentSolution",
-        "title": "MSBuild: Clean current solution"
+        "title": "MSBuild: Clean current solution",
+        "when": "fsharp.debuggerAvailable"
       },
       {
         "command": "fsharp.explorer.moveUp",
@@ -968,12 +978,12 @@
       {
         "command": "MSBuild.buildCurrent",
         "key": "ctrl\u002Balt\u002Bb",
-        "when": "fsharp.project.any \u0026\u0026 editorFocus \u0026\u0026 editorLangId == \u0027fsharp\u0027"
+        "when": "fsharp.project.any \u0026\u0026 editorFocus \u0026\u0026 editorLangId == \u0027fsharp\u0027 && fsharp.debuggerAvailable"
       },
       {
         "command": "MSBuild.buildCurrentSolution",
         "key": "ctrl\u002Balt\u002Bshift\u002Bb",
-        "when": "fsharp.project.any"
+        "when": "fsharp.project.any && fsharp.debuggerAvailable"
       },
       {
         "command": "fsharp.openInfoPanel",
@@ -1165,15 +1175,15 @@
         },
         {
           "command": "MSBuild.buildCurrent",
-          "when": "fsharp.project.any \u0026\u0026 editorLangId == \u0027fsharp\u0027"
+          "when": "fsharp.project.any \u0026\u0026 editorLangId == \u0027fsharp\u0027 && fsharp.debuggerAvailable"
         },
         {
           "command": "MSBuild.rebuildCurrent",
-          "when": "fsharp.project.any \u0026\u0026 editorLangId == \u0027fsharp\u0027"
+          "when": "fsharp.project.any \u0026\u0026 editorLangId == \u0027fsharp\u0027 && fsharp.debuggerAvailable"
         },
         {
           "command": "MSBuild.cleanCurrent",
-          "when": "fsharp.project.any \u0026\u0026 editorLangId == \u0027fsharp\u0027"
+          "when": "fsharp.project.any \u0026\u0026 editorLangId == \u0027fsharp\u0027 && fsharp.debuggerAvailable"
         },
         {
           "command": "fsharp.clearCache",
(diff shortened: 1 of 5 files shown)


src/Components/CSharpExtensionSupport.fs
```diff
@@ -0,0 +1,41 @@
+namespace Ionide.VSCode.FSharp
+
+open Fable.Import.VSCode.Vscode
+
+module CSharpExtension =
+
+    let private msCSharpExtensionName = "ms-vscode.csharp"
+    let private openvsixCSharpExtensionName = "ms-vscode.csharp"
+
+    let private resolvedCSharpExtensionName = msCSharpExtensionName
```
baronfel, line 10: Here is where we'd decide which extension ID to key off of - MS one or OSS one.

release/package.json
```diff
@@ -155,43 +155,53 @@
       },
       {
         "command": "MSBuild.buildCurrent",
-        "title": "MSBuild: Build current project"
+        "title": "MSBuild: Build current project",
+        "when": "fsharp.debuggerAvailable"
```
TheAngryByrd, line 159: YAK: should we nest `fsharp.debugger.available` or keep it flat?

src/Components/CSharpExtensionSupport.fs
```diff
@@ -0,0 +1,51 @@
+namespace Ionide.VSCode.FSharp
+
+open Fable.Import.VSCode.Vscode
+
+module CSharpExtension =
+
+    let private msCSharpExtensionName = "ms-vscode.csharp"
+    let private openvsixCSharpExtensionName = "ms-vscode.csharp"
```
TheAngryByrd, line 8: Where do these names come from? Are they different than the ids on the store pages?

release/package.json
baronfel: nested works for me - I'll do the global rename once I get all of the correct commands hooked up to the key

baronfel:
Ok, this is ready for other folks to poke at it. The build-related commands (restore, clean, build, rebuild) are always available, and the `run` commands are too. Only the debug commands are really disabled now, and the way we enforce this is by wrapping the call to generate the in-memory debug configurations. If no C# extension has been found, we generate no configurations. Otherwise we do our normal thing.

src/Components/CSharpExtensionSupport.fs
```diff
@@ -0,0 +1,86 @@
...
+                    commands.executeCommand ("extension.open", [| Some(box resolvedCSharpExtensionName) |])
+                    |> Promise.ofThenable
+                else
+                    Promise.empty)
+            |> Promise.catch (fun e ->
+                printfn $"Error installing C# extension: {Fable.Core.JS.JSON.stringify e}"
+                Promise.empty)
+            |> ignore<Fable.Core.JS.Promise<_>>
+
+            hasWarned <- true
```
baronfel, line 61: These two are designed to be cheap to call, so that components that rely on the debugger can match/call them inside processing loops naievely instead of having to be truly 'reactive'

src/Components/Debugger.fs
```diff
@@ -336,34 +336,39 @@ module Debugger =
...
+                                      | Some launchSettings ->
+                                          yield! configsForProject (p, launchSettings, buildTaskForProject)
+                                      | None -> ()
+                                      // emit a default configuration for this project if it is an executable
+                                      match defaultConfigForProject (p, buildTaskForProject) with
+                                      | Some p -> yield p
+                                      | None -> () ])
+
+                            return ResizeArray tasks
+                        }
```
baronfel, line 369: This is all the same logic, just wrapped in a 'if debugging exists then .. else' block

src/Components/MSBuild.fs
```diff
@@ -542,19 +542,6 @@ module MSBuild =
...
-        let typedMsbuildCmd f projOpt =
-            let p =
-                if JS.isDefined projOpt then
-                    Some(unbox<string> (projOpt))
-                else
-                    None
-
-            fun _ -> f p
-
         tasks.registerTaskProvider ("msbuild", msbuildBuildTaskProvider)
```
baronfel, line 545: These two functions were unused.

src/Ionide.FSharp.fsproj
```diff
@@ -1,57 +1,58 @@
...
-    <Compile Include="Components/InfoPanel.fs" />
-    <Compile Include="Components/CodeLensHelpers.fs" />
-    <Compile Include="Components/LineLens/PipelineHints.fs" />
-    <Compile Include="Components/SolutionExplorer.fs" />
-    <Compile Include="Components/TestExplorer.fs" />
-    <Compile Include="Components/InlayHints.fs" />
-    <Compile Include="fsharp.fs" />
-  </ItemGroup>
-  <Import Project="..\.paket\Paket.Restore.targets" />
+﻿<?xml version="1.0" encoding="utf-8"?>
```
baronfel, line 1: No actual change here - just line endings maybe?

baronfel:
If you're going to take this for a test drive, start by uninstalling C# and then reloading your environment. Debugging shouldn't work at all, but building/launching should.
Then install C# - you should get a notification that debugging has been enabled. Now debugging should work. 
If you remove C#, there's no actual change - because the extension isn't actually removed until you reload VSCode. This is the same flow as starting up and not finding C#, which you already covered.

baronfel:
@TheAngryByrd pointed out that the extension list query for "@category:debuggers F#" doesn't point directly to the C# extension (the vehicle for the functionality). We should reach out to that team to see what we can do.

Alternatively we could keep the commands enabled on our side, but have their implementations simply send the notification to install the C# extension directly.

TheAngryByrd:
We get some disappointing first generated `launch.json`.

```json
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": ".NET Core Launch (console)",
            "type": "coreclr",
            "request": "launch",
            "WARNING01": "*********************************************************************************",
            "WARNING02": "The C# extension was unable to automatically decode projects in the current",
            "WARNING03": "workspace to create a runnable launch.json file. A template launch.json file has",
            "WARNING04": "been created as a placeholder.",
            "WARNING05": "",
            "WARNING06": "If the server is currently unable to load your project, you can attempt to",
            "WARNING07": "resolve this by restoring any missing project dependencies (example: run 'dotnet",
            "WARNING08": "restore') and by fixing any reported errors from building the projects in your",
            "WARNING09": "workspace.",
            "WARNING10": "If this allows the server to now load your project then --",
            "WARNING11": "  * Delete this file",
            "WARNING12": "  * Open the Visual Studio Code command palette (View->Command Palette)",
            "WARNING13": "  * run the command: '.NET: Generate Assets for Build and Debug'.",
            "WARNING14": "",
            "WARNING15": "If your project requires a more complex launch configuration, you may wish to",
            "WARNING16": "delete this configuration and pick a different template using the 'Add",
            "WARNING17": "Configuration...' button at the bottom of this file.",
            "WARNING18": "*********************************************************************************",
            "preLaunchTask": "build",
            "program": "${workspaceFolder}/bin/Debug/<insert-target-framework-here>/<insert-project-name-here>.dll",
            "args": [],
            "cwd": "${workspaceFolder}",
            "console": "internalConsole",
            "stopAtEntry": false
        },
        {
            "name": ".NET Core Attach",
            "type": "coreclr",
            "request": "attach"
        }
    ]
}
```

Doubt this is something we can do but it will be a concern for newer people trying to debug a project.

TheAngryByrd:
Also trying to run a project from the Solution Explorer fails:

https://github.com/ionide/ionide-vscode-fsharp/assets/1490044/33257a3f-f4a8-42c6-b832-bc3af99d349e

Where it does seem to work in the current version.

TheAngryByrd:
Wonder if there's gonna be complications with https://github.com/ionide/ionide-vscode-fsharp/pull/1927/files

baronfel:
good spot - I'm assuming #1927 will go first. If it does then I'll do the same checks I've done here for the debugging commands.

baronfel:
> We get some disappointing first generated `launch.json`.
> 
> ```json
> {
>     // Use IntelliSense to learn about possible attributes.
>     // Hover to view descriptions of existing attributes.
>     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
>     "version": "0.2.0",
>     "configurations": [
>         {
>             "name": ".NET Core Launch (console)",
>             "type": "coreclr",
>             "request": "launch",
>             "WARNING01": "*********************************************************************************",
>             "WARNING02": "The C# extension was unable to automatically decode projects in the current",
>             "WARNING03": "workspace to create a runnable launch.json file. A template launch.json file has",
>             "WARNING04": "been created as a placeholder.",
>             "WARNING05": "",
>             "WARNING06": "If the server is currently unable to load your project, you can attempt to",
>             "WARNING07": "resolve this by restoring any missing project dependencies (example: run 'dotnet",
>             "WARNING08": "restore') and by fixing any reported errors from building the projects in your",
>             "WARNING09": "workspace.",
>             "WARNING10": "If this allows the server to now load your project then --",
>             "WARNING11": "  * Delete this file",
>             "WARNING12": "  * Open the Visual Studio Code command palette (View->Command Palette)",
>             "WARNING13": "  * run the command: '.NET: Generate Assets for Build and Debug'.",
>             "WARNING14": "",
>             "WARNING15": "If your project requires a more complex launch configuration, you may wish to",
>             "WARNING16": "delete this configuration and pick a different template using the 'Add",
>             "WARNING17": "Configuration...' button at the bottom of this file.",
>             "WARNING18": "*********************************************************************************",
>             "preLaunchTask": "build",
>             "program": "${workspaceFolder}/bin/Debug/<insert-target-framework-here>/<insert-project-name-here>.dll",
>             "args": [],
>             "cwd": "${workspaceFolder}",
>             "console": "internalConsole",
>             "stopAtEntry": false
>         },
>         {
>             "name": ".NET Core Attach",
>             "type": "coreclr",
>             "request": "attach"
>         }
>     ]
> }
> ```
> 
> Doubt this is something we can do but it will be a concern for newer people trying to debug a project.

I think if we can get the C# extension to add `fsharp` [here](https://github.com/dotnet/vscode-csharp/blob/main/package.json#L1934-L1939) then your search will work @TheAngryByrd. They are also trying to make an entirely new debug type labeled C# that is entirely dynamic (configuration for that is [here](https://github.com/dotnet/vscode-csharp/blob/main/package.json#L4336-L4365) that we might be able to copy from - this one is driving the 'f5 from nothing' experience for C# users.

PR #1908, opened 2023-08-10
Update CONTRIBUTING.md

### LINKED ISSUE #1907: Debugging docs: fsautocomplete source is no longer included on build
### Describe the bug
https://github.com/ionide/ionide-vscode-fsharp/blob/main/CONTRIBUTING.md#working-with-fsac
> Open FSAC from a new instance of VSCode from the directory: `paket-files/github.com/fsharp/FsAutoComplete`

However, fsautocomplete is now included as a package at `packages\fsac\fsautocomplete`. The source must be cloned separately. So steps 1 to 2 are outdated.

Happypig375:
Fixed #1907

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -144,7 +144,7 @@ Most Ionide bugs are actually from there. This repo only stores the glue code be
     ```bash
     git clone https://github.com/YOUR_GITHUB_USER/FsAutoComplete.git
     ```
-2. Follow the [FSAC build instructions](https://github.com/fsharp/FsAutoComplete#building-and-testing) and copy the dll output from the output log, it should be a path inside the FSAC repo: `src/FsAutoComplete/bin/Release/net6.0/fsautocomplete.dll`. Note that `Release` may be substituted with `Debug` if you build with that configuration instead. Also, note that `net6.0` may be replaced with another .NET version like `net7.0` as FsAutoComplete targets a newer .NET version.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -131,11 +131,21 @@ Note: Step 1 starts a new "Watch" task that rebuilds Ionide in the background wh
...
+    ```bash
+    git clone git@github.com:YOUR_GITHUB_USER/FsAutoComplete.git
+    ```
+    or if you don't use ssh:
+
+    ```bash
+    git clone https://github.com/YOUR_GITHUB_USER/FsAutoComplete.git
+    ```
+2. Follow the [FSAC build instructions](https://github.com/fsharp/FsAutoComplete#building-and-testing) and copy the dll output from the output log, it should be a path inside the FSAC repo ending with `src/FsAutoComplete/bin/Release/net6.0/fsautocomplete.dll`. Note that `Release` may be substituted with `Debug` if you build with that configuration instead. Also, note that `net6.0` may be replaced with another .NET version like `net7.0` as FsAutoComplete targets a newer .NET version.
+3. In the instance of VSCode that you have Ionide open, open settings, and find the section `FSharp > Fsac: Net Core Dll Path` and paste the output you copied from step 3.
```
baronfel, line 148: Note that in recent versions of Ionide (past couple months) you should just be able to point to the `bin/Release` part here and Ionide will handle negotiating which TFM to use. The exact path will also work, though.

Happypig375: Added it

baronfel:
Thanks @Happypig375!

PR #1913, opened 2023-08-21
Rewrite invokeMSBuild in terms of new invokeMSBuildWithCancel

farlee2121:
### WHAT
Unify `invokeMSBuild` and `invokeMSBuildWithCancel` to call back to the same core implementation.

### WHY
Because @baronfel [pointed out](https://github.com/ionide/ionide-vscode-fsharp/pull/1874#discussion_r1299238425) that I forgot to do it in my previous PR 

### HOW
`invokeMSBuild` creates a cancellation token so it can call `invokeMSBuildWithCancel`

### DIFF
--- a/src/Components/MSBuild.fs
+++ b/src/Components/MSBuild.fs
@@ -98,64 +98,8 @@ module MSBuild =
             |> Promise.ofThenable
 
     let invokeMSBuild project target =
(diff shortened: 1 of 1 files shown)


baronfel (approved):
Thanks!

PR #1920, opened 2023-08-31
Show package names only once

dawedawe:
### <samp>🤖 Generated by Copilot at b03d263</samp>

Fixed a bug in the solution explorer that caused duplicate package references to appear. Modified `packageRefs` in `src/Components/SolutionExplorer.fs` to filter out duplicates by name.

### <samp>🤖 Generated by Copilot at b03d263</samp>

> _`packageRefs` fixed_
> _No more duplicates shown in_
> _Solution explorer_

🐛🗑️🗃️

### WHY
Because there are multiple .dll files in the FSharp.Compiler.Service package, proj-info has two entries in the PackageReferences, which is fine. But here, we only want to show the package names.
image

### <samp>🤖 Generated by Copilot at b03d263</samp>

* Fix bug of duplicate package references in solution explorer by removing them before mapping to string ([link](https://github.com/ionide/ionide-vscode-fsharp/pull/1920/files?diff=unified&w=0#diff-13ae03b738fb9bcc3ffeb34994ecc2d80e815e98079c87f8e9c0c42fadd2a9d9L1046-R1048))

### DIFF
--- a/src/Components/SolutionExplorer.fs
+++ b/src/Components/SolutionExplorer.fs
@@ -1043,7 +1043,9 @@ module SolutionExplorer =
                     let projRefs = project.ProjectReferences |> Array.map (fun n -> n.ProjectFileName)
 
                     let packageRefs =
-                        project.PackageReferences |> Array.map (fun n -> n.Name + " " + n.Version)
+                        project.PackageReferences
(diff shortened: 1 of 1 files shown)


TheAngryByrd (commented):
💯 Thank you!

PR #1925, opened 2023-09-11
Trx parsing performance

farlee2121:
### WHAT
Dramatically improve trx result file parsing performance.

For ~2k unit tests, parse time went from ~260s (yikes) to <2s.

### WHY
While testing for #1922, I noticed large test projects had a significant delay between the test console showing a complete run and the explorer showing a complete run. A bit of experimenting showed that trx parsing was the source of the slowness.

I'm surprised I didn't notice this sooner. I thought the performance issues were in the build/run phase and got tunnel vision. Trx parsing was causing the non-linear slowdowns on large projects. 

### HOW

XPath has [known issues](https://github.com/goto100/xpath/issues/97) with performance for large documents.
So, I found a way to reduce the average selector scope. First I select a collection of test definitions or test results, then I query attributes relative to each test or result node. 
Test definitions and results are then zipped together by id. 

Overall, this requires a data model more closely reflecting the trx structure. It does slightly increase mapping complexity, but it drastically decreases document traversal via select statements.


baronfel (approved):
Absolutely lovely, and the changes to the parsing make sense to me overall. Good eye!

Issue #1923, opened 2023-09-04
Getting errors when trying to execute unit tests with special characters in the names via the test explorer


Numpsy:
### Describe the bug
I have a test project which uses Expecto and is using ```testTheory``` to create tests.
The generated test cases use the 'theory' data in the name, which resulted in some test cases whose name contains a ```=```.

The test explorer is displaying these test cases ok:

![image](https://github.com/ionide/ionide-vscode-fsharp/assets/1178570/3bbf8756-73f5-4b13-872d-2420e49ba233)

If I use the test explorer to run the whole test set, or just run the 'CLASSIFICATION' test on its own then the tests run ok.
However, if I try to run any of the tests whose name contains an = then I get this error in the test output

```
An exception occurred while invoking executor 'executor://yolodev/expecto': Incorrect format for TestCaseFilter Error: Invalid Condition 'FullyQualifiedName=TitusMetadataParser Tests.parse a user metadata string with no value.CLASSIFICATION='. Specify the correct format and try again. Note that the incorrect format can lead to no test getting executed.
```

I then tried running the xUnit based tests from [Avalonia.FuncUI](https://github.com/fsprojects/Avalonia.FuncUI) and found that there are similar issues with test cases that have '(' or ')' in the name, for example running 

https://github.com/fsprojects/Avalonia.FuncUI/blob/ffc04eb5f1f4865f48b1d0bac3c9cd9bb4bed33a/src/Avalonia.FuncUI.UnitTests/VirtualDom/VirtualDom.ModuleTests.fs#L151C19-L151C19

![image](https://github.com/ionide/ionide-vscode-fsharp/assets/1178570/f7df86f8-435e-4e86-8d2d-b32468bff480)

got me this error
```
[xUnit.net 00:00:00.47] Avalonia.FuncUI.UnitTests: Exception filtering tests: Incorrect format for TestCaseFilter Missing Operator '|' or '&'. Specify the correct format and try again. Note that the incorrect format can lead to no test getting executed.

No test matches the given testcase filter `(FullyQualifiedName=Avalonia.FuncUI.UnitTests.VirtualDom.ModuleTests.integration test 'VirtualDom.updateRoot' (from null to Button to TextBlock))` in s:\GitHubForks\MyFuncUI\src\Avalonia.FuncUI.UnitTests\bin\Debug\net6.0\Avalonia.FuncUI.UnitTests.dll
```

Fwiw, the text explorer in Visual Studio will run both of these cases.

[...]

PR #1926, opened 2023-09-12
Escape special characters when building test filters

Numpsy:
### WHAT
When building a test filter, escape any special characters present in test names, as per the vstest documentation at https://github.com/microsoft/vstest/blob/main/docs/filter.md, to avoid errors from the test runner when trying to run tests whose names contain such characters from the test explorer

### WHY
#1923 

### HOW
Based on the approach used by vstest at https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.ObjectModel/Utilities/FilterHelper.cs

### DIFF
--- a/src/Components/TestExplorer.fs
+++ b/src/Components/TestExplorer.fs
@@ -1060,10 +1060,32 @@ module Interactions =
         )
         |> Promise.ofThenable
 
+    let private filterEscapeCharacter = '\\'
+    let private filterSpecialCharacters = [| '\\'; '('; ')'; '&'; '|'; '='; '!'; '~' |]
+    let private filterSpecialCharactersSet = Set.ofArray filterSpecialCharacters
 
     let private buildFilterExpression (tests: TestItem array) =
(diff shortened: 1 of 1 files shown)


src/Components/TestExplorer.fs
```diff
@@ -1060,10 +1060,32 @@ module Interactions =
...
     let private buildFilterExpression (tests: TestItem array) =
+        // Escape any special characters in test names, as per https://github.com/microsoft/vstest/blob/main/docs/filter.md
+        let escapeFilterExpression (str: string) =
+
+            if str.IndexOfAny(filterSpecialCharacters) < 0 then
+                str
+            else
+                let builder = StringBuilder()
+
+                for i = 0 to str.Length do
```
Numpsy, line 1076: hmm, clicked submit and then thought 'hang on, that's one too many characters' :-(

TheAngryByrd (approved):
Thanks for this!

I don't know how I missed this in the last batch of merges.

PR #1927, opened 2023-09-12
Test debugging

farlee2121:
### WHAT
Add ability to debug unit tests from the Ionide Test Explorer.

https://github.com/ionide/ionide-vscode-fsharp/assets/2847259/7a9bbff8-56a0-4edb-b480-0a7276d30a79

Debug is also available from the gutter decoration
image

### HOW
Add a new test run profile for debugging. 
When the debugging profile is selected, 
1. set the right environment variable for vstest to return the processId (and wait for a debugger to attach)
2. parse the processId out of the output
3. attach to that process with a vscode attach profile
4. `workbench.action.debug.continue` to get past the stop on entry that can be confusing 

I tested to make sure it works across frameworks (Expecto, XUnit, etc) and when multiple projects are run.

### DIFF
--- a/src/Components/TestExplorer.fs
+++ b/src/Components/TestExplorer.fs
@@ -490,12 +490,19 @@ module DotnetCli =
         }
         |> ignore
 
+    type DebugTests =
+        | Debug
+        | NoDebug
+
+    module DebugTests =
+        let ofBool bool = if bool then Debug else NoDebug
+
     let private dotnetTest
         (cancellationToken: CancellationToken)
         (projectPath: string)
         (targetFramework: string)
         (trxOutputPath: string option)
-        (shouldDebug: bool)
+        (shouldDebug: DebugTests)
         (additionalArgs: string array)
         : JS.Promise<Node.ChildProcess.ExecError option * StandardOutput * StandardError> =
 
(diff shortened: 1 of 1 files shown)


src/Components/TestExplorer.fs
```diff
@@ -446,16 +452,50 @@ module DotnetCli =
...
+            |> unbox
+
+        let folder = workspace.workspaceFolders.Value.[0]
+
+        promise {
+            let! _ =
+                Vscode.debug.startDebugging (Some folder, U2.Case2 launchRequest)
+                |> Promise.ofThenable
+
+            do! Promise.sleep 2000
```
farlee2121, line 488: Have to wait or it'll continue before the debugger reaches the stop on entry point.
That'll leave the debugger in a confusing state where it shows it's attached but no breakpoints are hit and the breakpoints show as disabled

TheAngryByrd: Definitely leave a code comment for this one then.

farlee2121: Done. 
Also commented on the processId weirdness

TheAngryByrd (approved):
This works great! Thank you 🎸 

As a note for Expecto users, if you want to actually run only one test you'll need to update [YoloDev to 0.14.2](https://github.com/YoloDev/YoloDev.Expecto.TestSdk/releases/tag/YoloDev.Expecto.TestSdk-v0.14.2)

Issue #1924, opened 2023-09-07
No space after code before type lense is shown


peheje:
### Describe the bug
Before there was a space after the code before the type lense is shown.

Now this is gone:
<img width="295" alt="Screenshot 2023-09-07 at 20 25 10" src="https://github.com/ionide/ionide-vscode-fsharp/assets/7273366/0d33c49c-63a2-46d5-b6c9-82a6d76cb6ba">
Notice the `b//` used to be `b //`

This look especially wierd when font-ligatures is enabled:
<img width="199" alt="Screenshot 2023-09-07 at 20 26 32" src="https://github.com/ionide/ionide-vscode-fsharp/assets/7273366/7de64cd3-e983-4236-acd2-87eaf50c1660">

Notice the `=//` which should be `= //`

I'm on Ionide V 7.11.0 - 06.09.2023 and MacOS. But I think the issue was also present 1 version back, but just a hunch.

Numpsy:
Hopefully fixed in 7.12.0

PR #1928, opened 2023-09-13
Add a leading space to the default prefix in LineLens.config

Numpsy:
### WHAT
Add a leading space to the default prefix string in the shared line lens code.

### WHY
refs #1924 

Note that this only updates the default value and doesn't fix the actual issue, it just seemed reasonable to me for this default to have a leading space to match the one at https://github.com/ionide/ionide-vscode-fsharp/blob/61c4d71ef5b255a77aa368548530380174d39ac4/src/Components/LineLens/LineLens.fs#L35C53-L35C53

I'll do a test PR shortly for another possible change related to the issue

### HOW
It's just a space in the prefix string.

### DIFF
--- a/src/Components/LineLens/LineLensShared.fs
+++ b/src/Components/LineLens/LineLensShared.fs
@@ -186,7 +186,7 @@ type LineLens
     let decorationType =
         decorationType |> Option.defaultValue LineLensDecorations.decorationType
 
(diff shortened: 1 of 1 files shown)


TheAngryByrd (approved):
I wonder how this actually happened. Thanks for this!

PR #1931, opened 2023-09-15
Add keybinding to fsharp.explorer.renameFile in F# Solution Explorer

dawedawe:
### WHAT
copilot:summary

copilot:poem

copilot:emoji

### WHY
I thought this would be a quick one, but apparently I'm missing some background story here why this doesn't work with such a simple change. Does anyone know why?

### HOW
copilot:walkthrough

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -949,6 +949,11 @@
     ],
     "keybindings": [
       {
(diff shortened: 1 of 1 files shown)


dawedawe:
Mmh, no, seems to be something else. I'll do some more digging.

Issue #1935, opened 2023-09-25
Test explorer gets stuck at 'in progress' using xUnit `Theory` tests


marklam:
### Describe the bug
Test explorer gets stuck at 'in progress' using xUnit `Theory` tests. It seems to be specific to the Theory type test, where the test is run multiple times with different parameters.

The output window shows that the `dotnet test` run completed, but the tree still shows it as running.

The equivalent code in C# run via the C# dev kit doesn't get stuck.

### Steps to reproduce
(sample repo linked below)

Run an xUnit `Theory` test from the VSCode UI. The "Test Explorer" pane and "Test Results" tree both show in progress, although the "Test Results" window contains:
```text
WARNING: Overwriting results file: c:\Users\markl\AppData\Roaming\Code\User\workspaceStorage\947a947d1479406114d01bb6220c0a98\Ionide.Ionide-fsharp\TestResults\TestListProblemFs\TestListProblemFs.trx

Results File: c:\Users\markl\AppData\Roaming\Code\User\workspaceStorage\947a947d1479406114d01bb6220c0a98\Ionide.Ionide-fsharp\TestResults\TestListProblemFs\TestListProblemFs.trx



Passed!  - Failed:     0, Passed:     2, Skipped:     0, Total:     2, Duration: 3 ms - TestListProblemFs.dll (net6.0)
```

An equivalent C# project shows completion in the same windows.

### Link to sample reproduction
https://github.com/marklam/TestListProblem

### Expected behaviour
The tests should be marked as complete and passed

### Screenshots
![image](https://github.com/ionide/ionide-vscode-fsharp/assets/1962640/b73f15cc-ab90-40f7-b09b-00883fb5f30b)

## Machine info
 - OS: Windows
 - .NET SDK version: 6.0.414, 7.0.401
 - Ionide version: v7.12.3

TheAngryByrd:
@farlee2121

farlee2121:
Interesting. The issue is that xUnit doesn't include the theory data in the test method name in the test results file.
This means that all the theory tests try to register under the same name, and the explorer errors.

```xml
<!-- UnitTest.name includes the arguments -->
[...]

PR #1936, opened 2023-09-26
Bug xunit theory

farlee2121:
### WHAT
#1935 
Parameterized tests were never finishing for xUnit and MSTest.
This gets them completing properly.

### HOW
Piece together consistent fully qualified test names from different fields based on the test framework.

## Merge?

I'm not sure if this PR is worth merging yet.
It does get parameterized tests completing and status showing across xUnit, nUnit, MSTest, and expecto.

However, MSTest and xUnit still can't individually run the theory cases (some kind of filter issue), and the theory cases aren't nested under their shared method.
This makes it a impossible to run the theory test without running a larger grouping of tests.

### DIFF
--- a/src/Components/TestExplorer.fs
+++ b/src/Components/TestExplorer.fs
@@ -269,11 +269,11 @@ module TrxParser =
           Name: string }
 
     type UnitTest =
-        { Execution: Execution
+        { Name: string
+          Execution: Execution
           TestMethod: TestMethod }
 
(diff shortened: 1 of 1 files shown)


farlee2121:
There aren't any clear wholistic solutions to the issues with parameterized tests.
It might take me a while to get a good solution out there.
Therefore, I think it's worth getting this change out so users can at least run test suites including parameterized tests

PR #1943, opened 2023-10-05
Xunit theory filter

farlee2121:
### WHAT

#1935

#1936 got parameterized tests running as part of larger groups. 
This PR continues that work and allows individual cases to be run individually from
the test explorer (i.e. clicking the run button on a theory case in the test explorer)

Note that for MSTest, it will actually run all of the theory cases because there doesn't appear to be a way to filter to a single parameterized test case in MSTest. It'll still look right, but multiple cases will be run if they try to debug. 
All other frameworks (xUnit, nUnit, Expecto) can run individual parameterized cases and will debug as expected.

### HOW

Mostly through modifying how we build the `dotnet test` filter expressions based on which test framework the project uses.

### Nesting Deferred

I decided not to tackle nesting parameterized cases under a dedicated parent item in the test explorer (like Visual Studio does).
I'm still contemplating the options. Breaking the assumption that the explorer hierarchy matches the test name hierarchy could cause a lot of problems and complexity.

### DIFF
--- a/src/Components/TestExplorer.fs
+++ b/src/Components/TestExplorer.fs
@@ -75,6 +75,10 @@ module ListExt =
     let mapPartitioned f (left, right) =
         (left |> List.map f), (right |> List.map f)
 
+module Dict =
+    let tryGet (d: Collections.Generic.IDictionary<'key, 'value>) (key) : 'value option =
+        if d.ContainsKey(key) then Some d[key] else None
+
 module CancellationToken =
     let mergeTokens (tokens: CancellationToken list) =
         let tokenSource = vscode.CancellationTokenSource.Create()
@@ -227,6 +231,12 @@ module TestFrameworkId =
     [<Literal>]
     let MsTest = "MSTest"
 
+    [<Literal>]
+    let XUnit = "XUnit"
+
+    [<Literal>]
+    let Expecto = "Expecto"
+
 type TestResult =
     { FullTestName: string
       Outcome: TestResultOutcome
@@ -270,6 +280,10 @@ module TrxParser =
             Some TestFrameworkId.NUnit
         else if String.startWith "executor://mstest" adapterTypeName then
             Some TestFrameworkId.MsTest
+        else if String.startWith "executor://xunit" adapterTypeName then
+            Some TestFrameworkId.XUnit
+        else if String.startWith "executor://yolodev" adapterTypeName then
+            Some TestFrameworkId.Expecto
         else
             None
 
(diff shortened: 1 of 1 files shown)


src/Components/TestExplorer.fs
```diff
@@ -1067,7 +1081,7 @@ module TestDiscovery =
             trxTestsPerProject
             |> Array.map (fun (project, trxDefs) ->
                 let projectPath = ProjectPath.ofString project.Project
-                let heirarchy = TrxParser.inferHierarchy trxDefs
+                let hierarchy = TrxParser.inferHierarchy trxDefs
```
farlee2121, line 1084: just a spelling fix

src/Components/TestExplorer.fs
```diff
@@ -1180,17 +1193,32 @@ module Interactions =
 
         let testToFilterExpression (test: TestItem) =
             let fullTestName = TestItem.getFullName test.id
-            let fullName = escapeFilterExpression fullTestName
+            let escapedTestName = escapeFilterExpression fullTestName
```
farlee2121, line 1196: just a rename to indicate the intent of how the name changed

src/Components/TestExplorer.fs
```diff
@@ -1412,10 +1439,18 @@ module Interactions =
                 |> Option.map (fun p -> p.kind = TestRunProfileKind.Debug)
                 |> Option.defaultValue false
 
+            let hasIncludeFilter =
```
farlee2121, line 1442: This is actually a small unrelated fix.
Project-level items were constructing test filter clauses for all their top-level children.
But, that meant it wouldn't discover top-level tests if they weren't in the explorer yet.

The new behavior is to use no filter if just the top-level project is being run.
I can split this out if you'd prefer

src/Components/TestExplorer.fs
```diff
@@ -1180,17 +1193,32 @@ module Interactions =
...
+            let escapedTestName = escapeFilterExpression fullTestName
 
-            if fullName.Contains(" ") && test.TestFramework = TestFrameworkId.NUnit then
+            if escapedTestName.Contains(" ") && test.TestFramework = TestFrameworkId.NUnit then
                 // workaround for https://github.com/nunit/nunit3-vs-adapter/issues/876
                 // Potentially we are going to run multiple tests that match this filter
-                let testPart = fullName.Split(' ').[0]
+                let testPart = escapedTestName.Split(' ').[0]
                 $"(FullyQualifiedName~{testPart})"
+            else if test.TestFramework = TestFrameworkId.XUnit then
```
farlee2121, line 1203: This is the core of the change. It customizes the filter expression based on test framework.

src/Components/TestExplorer.fs
```diff
@@ -1544,7 +1579,12 @@ module Interactions =
                     logger.Error(message, buildFailures |> List.map ProjectPath.fromProject)
 
                 else
-                    let librariesCapableOfListOnlyDiscovery = set [ "Expecto"; "xunit.abstractions" ]
+                    let detectablePackageToFramework =
```
farlee2121, line 1582: xUnit and expecto can be discovered with `--list-tests` so we need to set the test framework in this discovery method too

TheAngryByrd (approved):
🎸 Thank you so much for your continued work on this!

Issue #1946, opened 2023-10-08
Test explorer: "Run test" works while "Debug test" fails to build FSharp compiler tests


Martin521:
### Describe the bug
In the situation detailed in "Steps to reproduce", I can run any test from the test explorer by clicking `Run Test`. However, if I click `Debug Test`, I get the error message shown below. Note that in both cases, I see a notification `Running MSBuild 'Build' target on '/workspaces/fsharp/tests/FSharp.Compiler.ComponentTests/FSharp.Compiler.ComponentTests.fsproj'`.

Error message in the `TEST RESULTS` window:
```
/workspaces/fsharp/FSharpBuild.Directory.Build.props(3,3): error : Could not resolve SDK "Microsoft.DotNet.Arcade.Sdk". Exactly one of the probing messages below indicates why we could not resolve the SDK. Investigate and resolve that message to correctly specify the SDK.
/workspaces/fsharp/FSharpBuild.Directory.Build.props(3,3): error :   SDK resolver "Microsoft.DotNet.MSBuildWorkloadSdkResolver" returned null.
/workspaces/fsharp/FSharpBuild.Directory.Build.props(3,3): error :   Failed to load NuGet settings. Required environment variable 'HOME' is not set. Try setting 'DOTNET_CLI_HOME' or 'HOME' and running the operation
/workspaces/fsharp/FSharpBuild.Directory.Build.props(3,3): error :   MSB4276: The default SDK resolver failed to resolve SDK "Microsoft.DotNet.Arcade.Sdk" because directory "/usr/share/dotnet/sdk/8.0.100-rc.1.23463.5/Sdks/Microsoft.DotNet.Arcade.Sdk/Sdk" did not exist.

/workspaces/fsharp/FSharpBuild.Directory.Build.props(3,31): error MSB4236: The SDK 'Microsoft.DotNet.Arcade.Sdk' specified could not be found. [/workspaces/fsharp/tests/FSharp.Compiler.ComponentTests/FSharp.Compiler.ComponentTests.fsproj]
```

### Steps to reproduce
1. Run `Dev Containers: Clone Repository in Named Container volume ...` and clone the dotnet/fsharp repository.
2. Wait for the dev container to build and the test explorer to discover the tests
3. Run any test in FSharp.Compiler.ComponentTests (=> succeeds)
4. Debug a test (=> above error message)

### Link to sample reproduction
[FSharp](https://github.com/dotnet/fsharp)

### Expected behaviour
[...]

PR #1951, opened 2023-10-19
Forward env variables in test debugging

farlee2121:
### WHAT
Ensure environment variables are forwarded when debugging tests.

### WHY
Issue #1946 revealed that the environment variables were not the same between debug and non-debug executions of a test.
This causes inconsistent behavior for tests that access environment variables. 

The cause of this bug was accidental override of the whole environment variable collection in order to set the debug flag for vstest.


TheAngryByrd (approved):
🎸 Thank you!

Issue #1653, opened 2021-12-30
How to get localisation disabled?


WhiteBlackGoose:
Hello!

First of all, great extension. Thanks for it!

Now, I'm quite new to VSC. I set the display language to English in it, but some things are still displayed not in English:
![image](https://user-images.githubusercontent.com/31178401/147763150-aff8d08e-b988-482a-86e8-9059d3e2a08c.png)

Is there a way to disable the localisation for the extension?

Sorry if it's not a good place for questions, I don't know where to ask otherwise.

baronfel:
that's really interesting, if you enable the LSP server logging (set `"FSharp.server.trace": "verbose"` in your settings.json) and reload the project, can you look in the F# output logs (there's an output panel with the outputs from all your extensions in the Command Palette), can you look for the request/response pairs for the `fsharp/signatureData` requests there?

baronfel:
The data is coming directly from the `fsharp/pipelineHint` from FSAC, so we'll need to make a fix there first.

baronfel:
It's likely coming from the [`getGenerics`](https://github.com/fsharp/FsAutoComplete/blob/8c7bfa815f195ea54b0b033a2c448ac1a24295a3/src/FsAutoComplete.Core/Commands.fs#L1588-L1603) function in the pipeline hints command. This extracts out the 'type mappings' section of the tooltip data, and the 'is' string in '<type parameter name> is <concrete instantiation>' is what's being translated here.

baronfel:
triage for later: may need to set/force FSAC to launch with a custom locale to debug this, due to the localization of the tooltips. @WhiteBlackGoose would you mind noting your locale here for later investigation?

PR #1952, opened 2023-10-21
fix and improve

Tangent-90:
### <samp>🤖 Generated by Copilot at 53f4abb</samp>

This pull request adds a new option for customizing external autocomplete, a feature for sending the current line to F# interactive, and a bug fix for the pipeline hints type signatures. It affects the files `release/package.json`, `src/Components/Fsi.fs`, and `src/Components/LineLens/PipelineHints.fs`.

### <samp>🤖 Generated by Copilot at 53f4abb</samp>

> _If you want to use F# in VS Code_
> _You might like this pull request's load_
> _It fixes `getSignature`_
> _Adds `fullNameExternalAutocomplete`_
> _And lets you send lines to `Fsi` mode_

🐛🛠️🚀

### WHY
1. add: new FSAC config fullNameExternalAutocomplete (fsharp/FsAutoComplete#1178) to package.json
2. fix: PipelineHint display wrong with localized type param output (wrong display examples like #1653), now it will correctly display hints without cutting too much to the type name, adding extra "1" or word "is" (like "是" in Chinese, "は" in Japanese).
3. improve: use correct line number when sending lines to fsi. This change makes the line numbers consistent in FSI and the editor, may making scripts easier to debug.

### <samp>🤖 Generated by Copilot at 53f4abb</samp>

*  Add a new configuration option for external autocomplete ([link](https://github.com/ionide/ionide-vscode-fsharp/pull/1952/files?diff=unified&w=0#diff-7902ed52e6f39559068a38e3c223d7f7d7d0b0977aee33d659db648092a62b15R528-R532))
* Support sending the current line to F# interactive terminal ([link](https://github.com/ionide/ionide-vscode-fsharp/pull/1952/files?diff=unified&w=0#diff-9bc76a29d86d81de8050e1137bb71027121b22b7df2f87cf5e05a8e43dd6b39bL322-R334), [link](https://github.com/ionide/ionide-vscode-fsharp/pull/1952/files?diff=unified&w=0#diff-9bc76a29d86d81de8050e1137bb71027121b22b7df2f87cf5e05a8e43dd6b39bL344-R355), [link](https://github.com/ionide/ionide-vscode-fsharp/pull/1952/files?diff=unified&w=0#diff-9bc76a29d86d81de8050e1137bb71027121b22b7df2f87cf5e05a8e43dd6b39bL492-R505))
* Fix type signature extraction for pipeline hints ([link](https://github.com/ionide/ionide-vscode-fsharp/pull/1952/files?diff=unified&w=0#diff-dbab5b128338ebe31e4a7f519fa49fb3c35e53a8ef76615c0f78afe15ada4da6L47-R53))
* Add a newline at the end of `release/package.json` ([link](https://github.com/ionide/ionide-vscode-fsharp/pull/1952/files?diff=unified&w=0#diff-7902ed52e6f39559068a38e3c223d7f7d7d0b0977aee33d659db648092a62b15L1774-R1779))

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -525,6 +525,11 @@
           "description": "Includes external (from unopened modules and namespaces) symbols in autocomplete",
           "type": "boolean"
         },
+        "FSharp.fullNameExternalAutocomplete": {
+          "default": false,
+          "description": "When selecting an external symbols in autocomplete, insert the full name to the editor rather than open its module/namespace. Also allow filtering suggestions by typing its full name. \n\n Requires `FSharp.externalAutocomplete` enabled.",
+          "type": "boolean"
+        },
         "FSharp.fsac.attachDebugger": {
           "default": false,
           "description": "Appends the \u0027--attachdebugger\u0027 argument to fsac, this will allow you to attach a debugger.",
--- a/src/Components/Fsi.fs
+++ b/src/Components/Fsi.fs
@@ -319,17 +319,19 @@ module Fsi =
 
     let mutable lastCd: string option = None
     let mutable lastCurrentFile: string option = None
+    let mutable lastCurrentLine: int option = None
 
     let private sendCd (terminal: Terminal) (textEditor: TextEditor option) =
-        let file, dir =
+        let file, dir, line =
             match textEditor with
             | Some textEditor ->
                 let file = textEditor.document.fileName
                 let dir = node.path.dirname file
-                file, dir
+                let line = int textEditor.selection.start.line + 1
+                file, dir, line
             | None ->
                 let dir = workspace.rootPath.Value
-                node.path.join (dir, "tmp.fsx"), dir
+                node.path.join (dir, "tmp.fsx"), dir, 1
 
         match lastCd with
         // Same dir as last time, no need to send it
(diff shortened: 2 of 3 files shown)


baronfel:
Hi, thanks for this! The fixes for 1. and 3. make sense to me - can you clarify on what the problem for 2. was? I'm having a hard time understanding what the minimal version of the problem is.

Tangent-90:
> Hi, thanks for this! The fixes for 1. and 3. make sense to me - can you clarify on what the problem for 2. was? I'm having a hard time understanding what the minimal version of the problem is.

When displaying pipeline hints, it will first find the type parameters of operator (|>), and then extract type names. The problem is that the original way to extract names is based on English (as it will first find the word "is"), and will reduced wrong output when compiler doesn't give an English text. 

For example, in this code snippet
``` fsharp
1
|> (+) 2
```
F# compiler will give " 'T1 is int " and " 'U is int " under English environment, and will extract type names correctly to "int". 
But in a Chinese environment, the results of compiler will be " 'T1 是 int " and " 'U 是 int ", both doesn't contains "is", so will display hints like "1 是 int " and " 是 int ". And when the type name contains "is", it will take the word after the "is" as the output.

Tangent-90:
As all the localized "is" listed in text of F# compiler is a single word (like [this](https://github.com/dotnet/fsharp/blob/12c2a400b078532f6d20c37351d62820d9ef1fab/src/Compiler/xlf/FSComp.txt.zh-Hans.xlf#L8275C1-L8279C20), and type parameters text is all like "'T is something", so we can use a regex to extract the type name out.

baronfel:
Also, unrelated to this PR - would you suggest that Ionide itself support localizations for strings that we have or emit? Like the settings descriptions, etc? I've recently set up localization for another VSCode extension and it's on my mind.

baronfel (approved):

baronfel:
Note that we'll need to publish and update FSAC before this setting will have any effect, but we can do that today so that we can push this out :)

Tangent-90:
> Ok, that makes sense. It is fair to say that a better FSAC endpoint targeted for this use case would make this task easier/less variable for non-English locales?

What does this exactly mean?

> Also, unrelated to this PR - would you suggest that Ionide itself support localizations for strings that we have or emit? Like the settings descriptions, etc? I've recently set up localization for another VSCode extension and it's on my mind.

if it's not difficult to achieve, then it is a good idea.

baronfel:
> >    Ok, that makes sense. It is fair to say that a better FSAC endpoint targeted for this use case would make this task easier/less variable for non-English locales?
>
> What does this exactly mean?

All I mean here is that I think it would be useful for FSAC to have an API specifically for powering pipeline hints (possibly using the same infrastructure as other inlay hints?) and let FSAC use a more structured form of the data instead of the string manipulation that currently happens in Ionide.

PR #1955, opened 2023-11-01
[WIP] Allow selecting fsproj files into workspace

purkhusid:
### WHAT
copilot:summary

copilot:poem

copilot:emoji

### WHY
When working in a monorepo with multiple independent projects it is not feasible to load all the projects in the repository. This change allows the user to select the projects that they are working on and only load them.

### HOW
copilot:walkthrough

### DIFF
--- a/src/Components/MSBuild.fs
+++ b/src/Components/MSBuild.fs
@@ -399,6 +399,7 @@ module MSBuild =
     let buildTaskListForSolution (s: WorkspacePeekFound) : JS.Promise<MSBuildTask seq> =
         promise {
             match s with
+            | WorkspacePeekFound.Fsproj _ -> return Seq.empty
             | WorkspacePeekFound.Directory _ -> return Seq.empty
             | WorkspacePeekFound.Solution({ Path = p }) ->
                 let! dotnet = dotnetBinary ()
--- a/src/Components/SolutionExplorer.fs
+++ b/src/Components/SolutionExplorer.fs
@@ -262,6 +262,11 @@ module SolutionExplorer =
             let result = Workspace items
             setParentRefs items result
             result
+        | WorkspacePeekFound.Fsproj proj ->
+            let result = getProjItem proj.Fsproj
+            let root = Workspace [ result ]
+            setParentRef result root
+            root
 
     let private getSolution () =
         Project.getLoadedSolution () |> Option.map getSolutionModel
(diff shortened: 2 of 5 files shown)


purkhusid:
@baronfel  @TheAngryByrd This is still WIP but I wanted to check if this was a feature that would be accepted. 

The defaults would have to be rethough a bit since Ionide always loads the available sln file instead of prompting the user.

I was also thinking about adding a new command for selecting multiple fsprojs, and then an configuration option that allows the user to set their default fsprojs.

baronfel:
There's a lot of overlap here between an already-existing `slnf` (solution filter) feature of MSBuild, but I don't see anything inherently wrong with the "open project + direct dependencies" scenario. I think as soon as you get to wanting to open "these N projects and their dependencies" I would look strongly at that format.

purkhusid:
I thought about going the slnf route but the downside of that is that you also need to maintain an actual solution file which contains all the projects. The main motivator for me is to make it as frictionless as possible to open a repo and loading only the projects that you need for the work you are about to be doing.

How I envisioned it is that a developer would do the following:
1. Open the repo folder in VS Code
2. If they are about to do work on F#(Could be a polyglot monorepo) they would CTRL + P -> F#: Load Projects
   * Would pop up a multiselect list that contains all the F# projects in the repo
3. They would select the projects they are interested in
4. Ionide would load the projects
5. Start coding

This workflow is really nice because you can configure your workspace without having to mess with any files at.
If you think we could make the slnf route as nice as the workflow I envisioned above then I would love to hear your ideas.

purkhusid:
There is also one other way to make the dev experience even easier and that would be:
1. Dev opens repo folder in VS Code
2. If they are about to do work on F# they just open the F# file they are going to work on
3. Ionide would find out what fsproj the opened file belongs to and would load it

This would be my ideal workflow but I think it's also quite a bit more complex to implement and that's why I went the route that is in the PR. If you think that lazy loading the projects like this would be possible then I think I'll have a go at it if you could point me in the right direction.

purkhusid:
I looked a bit into the lazy loading of projects and it seems like the main blocker for doing it is that we can't tell what fsproj an .fs file belongs to without calling `workspaceLoad` for the project. Do you have any suggestions on a way to map fsproj -> .fs files without doing a full load of the workspace?

baronfel:
In case you haven't seen it, [slngen](https://microsoft.github.io/slngen/) is some prior art in this space.

PR #1975, opened 2024-01-04
Fix egregious and unidiomatic equality syntax in the FSI watcher

baronfel:


src/Components/Fsi.fs
```diff
@@ -291,7 +291,7 @@ module Fsi =
                                 let hint =
                                     vscode.InlayHint.Create(
                                         line.range.``end``,
-                                        !!(" == " + value),
+                                        !!(" = " + value),
```
isaacabraham, line 294: I'm even wondering if the `" = "` is required at all. Oftentimes this is going to be directly alongside something of the form `let x =` so you already have the `=` there anyway.

PR #1989, opened 2024-03-05
Associate `.fantomasignore` with `ignore` files

MangelMaxime:
Doing so allows us to have nice icon for `.fantomasignore` file and also syntax highlighting, etc.

CleanShot 2024-03-05 at 19 04 28


baronfel:
Nice!

Issue #1997, opened 2024-03-30
New tests not detected


nojaf:
### Describe the bug

When adding a new unit test in an existing file, the new test does not get picked up.

### Steps to reproduce

- Try cloning https://github.com/fsprojects/fantomas
- Open project and file `src\Fantomas.Core.Tests\Stroustrup\SynBindingFunctionExpressionTests.fs`
- Wait until you can see test icon for existing tests
- add a new test like `[<Test>] let foo () = ()`

### Expected behaviour

The new test gets picked up.

### Screenshots

![image](https://github.com/ionide/ionide-vscode-fsharp/assets/2621499/a407e612-93a6-42e3-b028-135de31d371a)

## Machine info
 - OS: Windows
 - .NET SDK version: 8.0.202
 - Ionide version: v7.18.2

### Additional context

I noticed @farlee2121 changed the way unit tests are detected in https://github.com/ionide/ionide-vscode-fsharp/pull/1874. There no longer seems to be any notification coming from the LSP server.

Is the file being watched for changes? I can't really find any evidence for this in the current code.

farlee2121:
It still uses the same LSP code analysis notifications for live updates as before.
It now [routes through a mailbox](https://github.com/ionide/ionide-vscode-fsharp/blob/b0d2ebe13b497a34c7e475d8c65543d20c47fffb/src/Components/TestExplorer.fs#L1827) to resolve concurrency issues .
The main handler is [onTestsDiscoveredInCode](https://github.com/ionide/ionide-vscode-fsharp/blob/b0d2ebe13b497a34c7e475d8c65543d20c47fffb/src/Components/TestExplorer.fs#L1726)

nojaf:
Hmm, when attaching to FSAC, I see that I get an exception from:

![image](https://github.com/ionide/ionide-vscode-fsharp/assets/2621499/7df1ef31-69f7-495f-b81b-f7e06fd69cd1)

![image](https://github.com/ionide/ionide-vscode-fsharp/assets/2621499/a5bb4bca-400c-4e11-8537-33a45710b880)

The test is found but I suspect the notification is not getting to the client.

@baronfel, shot in the dark: is `FSharpSignatureData` trying to resolve a symbol with cached checking information? As it is failing for newly typed code that would be my guess.

PR #2000, opened 2024-04-05
Add test if not yet present in dictionary.

nojaf:
@farlee2121 I could be wrong but this seemed to help to show the new tests in my project.

Reasoning: the newly added test won't be in the existing dictionary and the check needs to be negated.
I could be wrong, please let me know if this makes sense.

### DIFF
--- a/src/Components/TestExplorer.fs
+++ b/src/Components/TestExplorer.fs
@@ -1048,7 +1048,7 @@ module TestDiscovery =
             removed |> Array.map TestItem.getId |> Array.iter targetCollection.delete
 
             added
-            |> Array.filter (TestItem.getId >> isKnownDisplacedFragment)
+            |> Array.filter (TestItem.getId >> isKnownDisplacedFragment >> not)
             |> Array.iter targetCollection.add
 
             unchanged


nojaf:
Fixes https://github.com/ionide/ionide-vscode-fsharp/issues/1997

farlee2121:
If I remember right, that condition is adding tests to the explorer when we've previously been able to match a non-standard location (e.g. a referenced, but not directly nested testList in Expecto). The displaced fragment cache is constructed beforehand isn't being modified in `mergeCodeUpdates`.

I ran some testing and something is definitely off about the explore behavior on code update. Renaming methods is causing them to disappear. I am able to add and remove test methods and see live updates though.

nojaf:
Similarly, when I set `"FSharp.TestExplorer.AutoDiscoverTestsOnLoad"` to `false`, my tests are no longer discovered at all. Even though I see the server sending the notifications with the correct information.

The reason I want `FSharp.TestExplorer.AutoDiscoverTestsOnLoad` to be false is that it runs all unit tests when I open my solution. I get why this is useful in some cases but for the Fantomas tests, all tests can be discovered via the AST quite accurately.

farlee2121:
If I'm understanding right, you want to discover tests on load via code analysis alone and not running tests.

That works against the base assumptions of the test explorer. A lot of the work I did was to re-orient the test explorer around test results since the code analysis results were causing flakiness (i.e. incorrectly located tests which broke the explorer).

nojaf:
> If I'm understanding right, you want to discover tests on load via code analysis alone and not running tests.

Yes, this used to work fine for me when I originally ported the test explorer code from Neptune.
Right now I'm a bit annoyed that all my tests are needlessly run when I open vscode (with setting on) or when I use the refresh action (with setting off).

The notifications work fine for me, this should also be possible to work with.

farlee2121:
My concern is that building from code analysis results would re-introduce the issues outlined in #1874.  Test results aren't always the fastest, but they are a consistently reliable method for constructing the test explorer.

I'm open to suggestions.

nojaf:
I would go for some middle ground here.
Not all projects face the issue of #1874, so there should be an opt-out.
In some codebases, it really is that simple that every `[<Test>]` equals a runnable test entry.

My suggestion would be to always process the incoming notifications. This seems to be off when `"FSharp.TestExplorer.AutoDiscoverTestsOnLoad": false` I suspect. And I would assume the server sends them anyway.

farlee2121:
Building the test explorer from code analysis isn't a matter of leaving it on when result-based updates are off, it'd require a separate mode. Currently, code analysis is not used as a primary truth source. It relies on the baseline provided by test results to prevent the unusable states described in #1874 (i.e. where run all doesn't work anymore).

Though, the explorer now should be able to run when non-existent tests are selected and be refreshed from test results if it gets off. Perhaps a "trust all code analysis" mode wouldn't get the explorer into an unrecoverable state.

farlee2121:
I've been digging a bit to understand how Visual Studio manages test location discovery.
I'm still getting a handle on it, but they seem to manage the whole discovery process through the Adapter Extensibility model.

Maybe that's the way forward here too? Then we're capable of anything VS is capable of and we can offload discovery to the adapters and hopefully some of the orchestration to vstest.

I'll keep digging.

farlee2121:
I've not gotten as much research into this as I'd hoped, but I did poke around VS Test and found that this is the package meant to smooth integration for editors: [Microsot.TestPlatform.TranslationLayer](https://www.nuget.org/packages/Microsoft.TestPlatform.TranslationLayer). 
They have an [example](https://github.com/Microsoft/vstest/tree/main/samples/Microsoft.TestPlatform.TranslationLayer.E2ETest) and a [playground for internal developers](https://github.com/microsoft/vstest/blob/main/playground/TestPlatform.Playground/Program.cs) that demonstrate it's use.

PR #2016, opened 2024-06-17
update FSAC and add support for new FSI options

baronfel:
Also includes grammar fixes and fixes for an accidental double registration of the rename file command.


src/Components/Fsi.fs
```diff
@@ -359,9 +359,10 @@ module Fsi =
 
         let parms =
             let fsiParams =
-                "FSharp.fsiExtraParameters"
-                |> Configuration.get Array.empty<string>
-                |> List.ofArray
+                Array.append
+                    (Configuration.get Array.empty<string> "FSharp.fsiExtraParameters")
+                    (Configuration.get Array.empty<string> "FSharp.FSIExtraInteractiveParameters")
```
baronfel, line 364: @TheAngryByrd quick pointer here - to prevent user churn while the new FSI parameter names are rolled out I've joined the two parameter arrays. This is documented in the notes for the new configuration settings at least, what do you think?

baronfel:
Alright, this looks good! Will kick off a new release.

Issue #1859, opened 2023-04-03
error FS0193: Operation is not supported on this platform.


jkone27:
### Describe the bug
I have a compiling script in F#

### Steps to reproduce
not clear, but sometimes happens in interactive while running the code with ALT-ENTER, not able to reproduce consistently yet.

### Expected behaviour
no errors as the code is valid F# code

## Machine info
 - OS: Mac
 - .NET SDK version: 7.0.202
 - Ionide version: 7.5.2

7enderhead:
I see this phenomenon, too. Linux Mint 20.3, VS Code 1.78.2, Ionide 7.5.4. The following code snippet, sent to an FSI terminal, works with dotnet 6.0.488 installed, but not with 7.0.302 or 8.0.100-preview.4.23260.5:

```fsharp
(* ~\~ language=FSharp filename=src/dnlib-testbed.fsx *)
(* ~\~ begin <<lit/dnlib-testbed.md|src/dnlib-testbed.fsx>>[init] *)
#if INTERACTIVE
(* ~\~ begin <<lit/dnlib-testbed.md|import-nugets>>[init] *)
#r "nuget:Microsoft.CodeAnalysis.CSharp" 
(* ~\~ end *)
(* ~\~ begin <<lit/dnlib-testbed.md|import-nugets>>[1] *)
#r "nuget:dnlib"
(* ~\~ end *)
(* ~\~ begin <<lit/dnlib-testbed.md|import-nugets>>[2] *)
#r "nuget:spkl.Diffs"
(* ~\~ end *)
(* ~\~ begin <<lit/dnlib-testbed.md|import-nugets>>[3] *)
#r "nuget:xunit"
#r "nuget:Unquote"
(* ~\~ end *)
#endif

(* ~\~ begin <<lit/dnlib-testbed.md|import-ms-codeanalysis>>[init] *)
open Microsoft.CodeAnalysis
open Microsoft.CodeAnalysis.CSharp
(* ~\~ end *)
(* ~\~ begin <<lit/dnlib-testbed.md|import-dnlib>>[init] *)
open dnlib.DotNet
open dnlib.DotNet.Emit
(* ~\~ end *)
(* ~\~ begin <<lit/dnlib-testbed.md|import-spkl-diffs>>[init] *)
open spkl.Diffs
(* ~\~ end *)
(* ~\~ begin <<lit/dnlib-testbed.md|import-dotnet>>[init] *)
open System
open System.IO
[...]

PR #2049, opened 2024-11-10
(fix)1859: FSI send ignore comments #1859

jkone27:
fixes issue with comments too long , partial fix for general issue 1859, to avoid `FS0193: operation not supported`

image

but comments can be safely ignored and skip FSI send, even if very long

image

### DIFF
--- a/src/Components/Fsi.fs
+++ b/src/Components/Fsi.fs
@@ -494,19 +494,40 @@ module Fsi =
             return terminal
         }
 
+    let detectTerminalSize (terminal: Terminal) =
+        promise {
+            let! terminalId = Promise.ofThenable terminal.processId
+
+            match terminalId with
+            | Some id ->
+                let! terminalSize = execCommand "stty" [ "size"; "-F"; sprintf "/proc/%d/fd/0" id ]
+                let rows, cols = terminalSize.Split(' ') |> Array.map int |> fun arr -> arr.[0], arr.[1]
+                return Some(rows, cols)
+            | None -> return None
+        }
+
     let private send (terminal: Terminal) (msg: string) =
         let msgWithNewline = msg + (if msg.Contains "//" then "\n" else "") + ";;\n" // TODO: Useful ??
         let linesCount = msgWithNewline |> Seq.filter ((=) '\n') |> Seq.length
 
         promise {
-            terminal.sendText (msgWithNewline, false)
-            lastSelectionSent <- Some msg
+            let! terminalSize = detectTerminalSize terminal
 
-            lastCurrentLine <-
-                match lastCurrentLine with
-                | Some line -> line + linesCount
-                | _ -> linesCount + 1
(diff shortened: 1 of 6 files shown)


src/Components/Fsi.fs
```diff
@@ -494,13 +496,27 @@ module Fsi =
             return terminal
         }
 
+    let private commentRegex =
+        Regex(@"\/{2,3}.+\S+", RegexOptions.ECMAScript)
+
+    let private multilineCommentRegex =
+        Regex(@"\(\*[\s\S]*?\*\)", RegexOptions.ECMAScript)
```
vzarytovskii, line 503: Should these be compiled?

vzarytovskii (commented):
Are there any interactions with line numbers and fsi in ionide? I.e. if we strip comments, and then repl will emit a diagnostic, is it navigable back to editor? In this case, it needs to somehow be compensated.

src/Components/Fsi.fs
baronfel: @vzarytovskii these are compiled to JS regexes, and I don't think there's a 'compiled' equivalent on that runtime.

baronfel:
@vzarytovskii you make a good point about the compensation - right now the closest thing we have to an 'in-editor' link is the fact that some terminals will look for things that look like a file path/line/column and will auto-link to it. VSCode does do this, so the line numbers might be a bit off.

This is something that would be fixed for real if VSCode managed FSI sessions interactively via the FCS APIs instead of by 'driving' `dotnet fsi` instances.

src/Components/Fsi.fs
vzarytovskii: Ah, I see

vzarytovskii:
> @vzarytovskii you make a good point about the compensation - right now the closest thing we have to an 'in-editor' link is the fact that some terminals will look for things that look like a file path/line/column and will auto-link to it. VSCode does do this, so the line numbers might be a bit off.
> 
> This is something that would be fixed for real if Ionide managed FSI sessions interactively via the FCS APIs instead of by 'driving' `dotnet fsi` instances.

The "cheapest" solution might be is to not remove comments, but strip them to just `//`, it will preserve lines, but no need for maintaining offsets.

src/Components/Fsi.fs
```diff
@@ -494,13 +496,27 @@ module Fsi =
             return terminal
         }
 
+    let private commentRegex =
+        Regex(@"\/{2,3}.+\S+", RegexOptions.ECMAScript)
```
vzarytovskii, line 500: This won't match the following:
```
//a
<any number of spaces here>//a<any number of spaces here>
```
First is quite short, so I suppose it's not an issue, second one might cause some issues?

This will match the following:
```
let foo = "// asdasdasdasdasdasd"
```

Which probably is not desired.

Second regex probably has some similar issues.

PR #2055, opened 2024-12-06
Update watcher.fsx to support nullness

isaacabraham:
I have zero experience on the Ionide codebase, but saw that VSCode was giving nullness warnings on starting FSI e.g.

```
ionide.ionide-fsharp-7.22.0\watcher\watcher.fsx(48,34): warning FS3261: Nullness warning: The types 'System.Reflection.MemberInfo' and 'System.Reflection.MemberInfo | null' do not have compatible nullability.
```

This PR removes the warnings, but someone should review them to make sure that there hasn't been any semantic change in behaviour.

I also saw one apparently unnecessary transformation from Array to List to Seq which I have removed, for performance purposes.


baronfel (approved):

baronfel:
Thanks @isaacabraham!

PR #2059, opened 2025-02-12
fixed error introduced in 5fcad29 that breaks the evaluation of values

enzian:
When evaluating expressions in FSI `watcher.fsx` loaded, evaluation of expressions yielded only the name of the variable but not the value anymore. This behaviour seems to have been introduced by a string formatting change in 5fcad29.
The change on the respective line was rolled back and tested - evaluation now works as before the update to version 7.23.


enzian:
bad edit on my part, I'll resubmit when the fix works correctly!

PR #2079, opened 2025-04-06
Proof of concept using different base paths

TheAngryByrd:


src/Components/MSBuild.fs
```diff
@@ -24,7 +24,7 @@ module MSBuild =
             | Ok msbuild -> Promise.lift msbuild
             | Error msg -> Promise.reject (exn msg))
 
-    let invokeMSBuildWithCancel project target (cancellationToken: CancellationToken) =
+    let invokeMSBuildWithCancel project target (cancellationToken: CancellationToken) args =
```
TheAngryByrd, line 27: We're doing the classic calling `msbuild t:Build` here instead of `dotnet build` which can cause issues with needing to restore. This is currently in a hacky state so we can discuss options.

src/Components/TestExplorer.fs
```diff
@@ -94,10 +94,21 @@ module CancellationToken =
...
+
+
 type TestId = string
 type ProjectPath = string
 type TargetFramework = string
 
+module Project =
+    let testPathSubDir = ".ionide-test"
+    let getOutputPaths () =
+        let objOutputPath = node.path.join [|  "obj" ; testPathSubDir  ; node.path.sep|]
```
TheAngryByrd, line 106: Using relative paths means it can build each respective project in its own subfolder (Real lib and test project), otherwise, things like project.assets.json will collide. . Though I'm not entirely sure this is the correct approach.

TheAngryByrd:
cc @farlee2121

src/Components/TestExplorer.fs
```diff
@@ -94,10 +94,21 @@ module CancellationToken =
...
         tokenSource.token
 
+
+
 type TestId = string
 type ProjectPath = string
 type TargetFramework = string
 
+module Project =
+    let testPathSubDir = ".ionide-test"
```
baronfel, line 104: idea: consider using [`ExtensionContext.storageUri`](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext) from the `activate` call to provide our workspace-unique state path?

src/Components/MSBuild.fs
baronfel: on thing to consider is that `/t:Build` calls the build target - and not all projects will have that target. Most _managed_ projects will, but what `dotnet build` and `msbuild` do by default is call the _default target_ so the most correct thing to do might be to allow specifying an optional target?

you can also do `dotnet build` with `--no-restore` if you want to skip the implicit restore.

src/Components/TestExplorer.fs
baronfel: using separate subfolders is the right move here IMO

baronfel: This is the same discussion that's going on for the `dotnet run foo.cs` work, and separate intermediate/output paths is what that effort has landed on as well.

Issue #2069, opened 2025-03-01
Setting TestingPlatformDotnetTestSupport makes tests non discoverable


OnurGumus:
### Describe the bug
There is a new post from [microsoft dev blogs]( https://devblogs.microsoft.com/dotnet/mtp-adoption-frameworks/?nsl_bypass_cache=5297fdc8a6815a8360ffe9b7112dcdec#comments)

They introduced `TestingPlatformDotnetTestSupport` switch in the project file. Once you set that true, tests aren't discoverable by ionide.
```
     <TestingPlatformDotnetTestSupport>true</TestingPlatformDotnetTestSupport>
```

To be fair I am not entirely sure what this new switch brings to table.

### Steps to reproduce
Create any test project then <TestingPlatformDotnetTestSupport>true</TestingPlatformDotnetTestSupport>

Ionide will complain no tests are discoverable.

### Expected behaviour
Ionide should discover the tests

## Machine info
 -  MacOS 15
 - .NET SDK version:9 
 - Ionide version: 7.25

TheAngryByrd:
cc @farlee2121

farlee2121:
I tested this, and found some significant problems.

First, Microsoft.Testing.Platform doesn't respect `<GenerateProgramFile>false</GenerateProgramFile>`. 

This means that using versions of YoloDev.Expecto.TestSdk which depend on Microsoft.Testing.Platform _can't_ define their own main method (and thus can't be run as a normal expecto project). This is true even if `EnableExpectoTestingPlatformIntegration` and `TestingPlatformDotnetTestSupport`  aren't enabled.

Second, Microsoft.Testing.Platform doesn't respect dotnet test flags like --logger and --list-tests.
This makes it incompatible with the way the Ionide test explorer currently works, which relies on --logger to discover tests and to report test results.

I confirmed that the --logger and --list-tests support is also missing from XUnit, not just YoloDev/Expecto.

In short, Microsoft.Testing.Platform won't work with Ionide short of rewriting test and result discovery or Microsoft.Testing.Platform better respecting `dotnet test` parameters.

TheAngryByrd:
Thanks for looking into this @farlee2121!

Youssef1313:
Hi @OnurGumus, @farlee2121, and @TheAngryByrd 

Let me clarify few points.

---

> First, Microsoft.Testing.Platform doesn't respect `<GenerateProgramFile>false</GenerateProgramFile>`.

[...]

PR #2082, opened 2025-05-03
Log standard out on passing tests

pblasucci:
This PR slightly modifies the code in `TestExplorer.fs`, such that if a TRX file contains output, it is emitted for passing tests. This is similar to the behavior of JetBrains Rider. Further, it facilitates having meaningful output when, eg, using FsCheck. 

For example, the following test:
```fsharp
[<Property>]
let ``test some stuff`` (PositiveInt value) =
    (0 < value) |> Prop.collect value
```
will print output like the following:
image

Happy to hear feedback on this (esp. from @farlee2121 and @TheAngryByrd). Thanks!

### DIFF
--- a/src/Components/TestExplorer.fs
+++ b/src/Components/TestExplorer.fs
@@ -240,6 +240,7 @@ module TestFrameworkId =
 type TestResult =
     { FullTestName: string
       Outcome: TestResultOutcome
+      Output: string option
       ErrorMessage: string option
       ErrorStackTrace: string option
       Expected: string option
@@ -312,13 +313,12 @@ module TrxParser =
         { Message: string option
           StackTrace: string option }
 
-    type Output = { ErrorInfo: ErrorInfo }
-
     type UnitTestResult =
         { ExecutionId: string
           Outcome: string
           Duration: TimeSpan
-          Output: Output }
+          Output: string option
+          ErrorInfo: ErrorInfo }
 
     type TestWithResult =
         { UnitTest: UnitTest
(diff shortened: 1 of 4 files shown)


src/Components/TestExplorer.fs
```diff
@@ -312,13 +313,12 @@ module TrxParser =
         { Message: string option
           StackTrace: string option }
 
-    type Output = { ErrorInfo: ErrorInfo }
```
farlee2121, line 315: UnitTestResult and it's children (like output) are made to reflect the structure of the Trx file.
To follow that pattern, the new data would be included in this Output object
```fs
type Output = 
  { 
    ErrorInfo: ErrorInfo 
    StdOut: string
  }
```

farlee2121:
I think that's a great feature add!

I left one comment on the intended structure of the TRX types.

Some related work: I'm working on moving us away from TRX parsing to using the [VsTest platform](https://www.nuget.org/packages/Microsoft.TestPlatform.TranslationLayer/17.14.0-preview-25107-01) and eventually the [Microsoft Testing Platform](https://github.com/ionide/ionide-vscode-fsharp/issues/2069).

I did some poking around to make sure we'll be able to support StdOut once the move from TRX to VsTest or MTP is done.
[MTP clearly supports standard out per test](https://github.com/microsoft/testfx/blob/1558cc3edf22e08938ffb30e2deafae0fd3b9d8a/docs/mstest-runner-protocol/001-protocol-intro.md?plain=1#L373), but I haven't been able to track down how it'd work with VsTest. It's got to be possible if the info is available in the trx file.

pblasucci:
> I think that's a great feature add!

Thanks! Happy to help. I've change the code per your suggestions.

> Someone else might have thoughts about the global.json and package.json changes.

For context, I needed the change to global.json in order to get things to build (without my having to go hunt down a specific older version of the SDK). However, I have no idea, really, about the packages.json file -- it's just something the tooling automatically changed for me (I really do _not_ understand all the Fable / JS / TS wizardry).

TheAngryByrd:
> global.json

It's such a pain, it's always wrong 😂 . Don't worry about it.

> package.json

looks to be pinning a version of yarn, which is probably fine.

TheAngryByrd (approved):
Gonna approve it. @pblasucci have any other changes before I merge and ship it?

PR #2088, opened 2025-05-21
Check for C# extension dependencies at runtime

### LINKED ISSUE #2087: Ionide and Cursor (C# Anysphere)
### Describe the bug

When trying to use Ionide in [Cursor AI code editor](https://www.cursor.com/) I get the error:

Cannot activate the 'Ionide for F#' extension because it depends on the 'C#' extension from 'Anysphere', which is not installed. Would you like to install the extension and reload the window?

However, the C# extension from Anysphere (id:anysphere.csharp) is already installed and installing it again will not help. The C# extension has the open-source netcoredbg debugger built in so have a feeling this should be able to work.

### Steps to reproduce

1. Open Cursor Editor
2. Activate the C# Anysphere extension
3. Activate Ionide extension
1. Open an F# file

### Screenshots

Image

### Machine infos
 - OS: Linux Pop!_OS 22.04 LTS
 - .NET SDK version: 9.0.300
 - Ionide version: 7.25.10

dbrattli:
## Why

We should not have a hard dependency on `ms-dotnettools.csharp` since that makes it impossible to use Ionide with forks os VScode e.g Cursor. This PR does the check at runtime instead adding recommended extensions to the Extension Pack tab instead.

Fixes #2087 

## How

- Check for valid C# extensions at runtime instead of in the manifest
- Suggest valid dependencies ast Extension Pack instead to allow for easy navigation to required C# extension
- Give error message if no valid C# extension can be found

### DIFF
--- a/release/package.json
+++ b/release/package.json
@@ -1831,8 +1831,12 @@
   "engines": {
     "vscode": "^0.10.0"
   },
+  "extensionPack": [
+    "ms-dotnettools.csharp",
+    "anyshere.csharp",
+    "muhammad-sammy.csharp"
+  ],
   "extensionDependencies": [
-    "ms-dotnettools.csharp"
   ],
   "homepage": "http://ionide.io",
   "icon": "images/logo.png",
--- a/release/package.json
+++ b/release/package.json
@@ -1833,7 +1833,7 @@
   },
   "extensionPack": [
     "ms-dotnettools.csharp",
-    "anyshere.csharp",
+    "anysphere.csharp",
     "muhammad-sammy.csharp"
   ],
   "extensionDependencies": [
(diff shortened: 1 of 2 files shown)


release/package.json
```diff
@@ -1831,8 +1831,12 @@
   "engines": {
     "vscode": "^0.10.0"
   },
+  "extensionPack": [
+    "ms-dotnettools.csharp",
+    "anyshere.csharp",
+    "muhammad-sammy.csharp"
+  ],
```
baronfel, line 1838: how does this present in the VSCode extension management UI?

dbrattli: This is how it looks for VScode and Cursor:

Screenshot 2025-05-30 at 15 53 25
Screenshot 2025-05-30 at 15 53 41

baronfel: perfect, thanks!

PR #2098, opened 2025-08-10
Fix tests disappearing from the explorer when their file is edited

farlee2121:


TheAngryByrd (approved):
Sorry for the late review, I was on vacation for a while. 

Thank you!

Issue #1941, opened 2023-10-04
Ship default keybindings for 'Move file up' and 'Move file down' in the project explorer


baronfel:
Users expect `Alt+Up/Down` to move files in a project based on experiences in other editors. We should ship this keybinding when in the context of a given F# project, like other extensions do:

![image](https://github.com/ionide/ionide-vscode-fsharp/assets/573979/e72e93ef-0cd4-440d-bcea-c6516b87393f)

PR #2105, opened 2025-09-14
add keybindings for move file up/down

pods:
Implementing #1941
Reordered `SolutionExplorer.activate` so that `createTreeView` is called before `registerCommand`.
This ensures that the command handler can access the model via `treeView.selection`.
When invoked by a keybinding, the model is not passed directly to the handler.


TheAngryByrd (approved):
Thank you!!

PR #2124, opened 2026-02-03
Add more Debug logs when activating features

64J0:
## Description

With this small PR, I'm adding more `Debug` logs when activating specific features. Perhaps this could be useful to identify whether some activation is hanging.

This is how the logs look like:

image


Krzysztof-Cieslak:
Thanks

